PR #5, opened 2021-11-23
Fixup sln file

pranavkm:


pranavkm:
I regened the sln file from scratch so the diffs might appear more significant.

dougbu (commented):
.sln changes look like something I should have done while rewriting our history.

But, what are the 52 .xlf files❔ I specifically avoided those files while doing patches based on your initial commit.

pranavkm:
The previous build was failing because it demanded the xlf be checked in (or a flag be configured):

> .packages/microsoft.dotnet.xlifftasks/1.0.0-beta.21518.1/build/Microsoft.DotNet.XliffTasks.targets(91,5): error : (NETCORE_ENGINEERING_TELEMETRY=Build) 'xlf/ComponentResources.cs.xlf' for 'ComponentResources.resx' does not exist. Run `msbuild /t:UpdateXlf` to update .xlf files or set UpdateXlfOnBuild=true to update them on every build, but note that it is strongly discouraged to set UpdateXlfOnBuild=true in official/CI build environments as they should not modify source code during the build.

dougbu (approved):
Once you get CI to pass, looks good

PR #8, opened 2021-11-29
Bump compiler version

pranavkm:
6.0.0 already shipped so we need a slightly higher version. Bumping it up to 6.1.0 to try and emulate Roslyn's versioning scheme.


dougbu (approved):
This works but why not bump `$(PatchVersion)` instead❔

PR #10, opened 2021-12-02
Stabilize versions

dougbu:


eng/Versions.props
```diff
@@ -4,12 +4,11 @@
     <MajorVersion>6</MajorVersion>
     <MinorVersion>1</MinorVersion>
     <PatchVersion>0</PatchVersion>
```
wtgodbe, line 6: 6.1.0?

wtgodbe (approved):

eng/Versions.props
dougbu: I didn't change this and defer to @pranavkm and @NTaylorMullen 

My suspicion is this version was chosen to avoid downgrades. I also suspect the exact choices don't matter much as long as we don't ship the packages publicly. If we _do_ want to ship them publicly, should instead move to `v6.0.2` for a future .NET 6.0.2 release.

dougbu: Oops, I _did_ change the versions and we _are_ at v6.0.2.

eng/Versions.props
```diff
@@ -4,11 +4,11 @@
     <MajorVersion>6</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>2</PatchVersion>
-    <PreReleaseVersionLabel>1</PreReleaseVersionLabel>
```
dougbu, line 7: such an ugly label in any case 😀

eng/Versions.props
```diff
@@ -4,11 +4,11 @@
...
+    <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
     <!--
         When StabilizePackageVersion is set to 'true', this branch will produce stable outputs for 'Shipping' packages
     -->
-    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
```
pranavkm, line 11: Looking at https://github.com/dotnet/sdk/pull/22864/files, it looks the build now produces stable builds of these packages. We're not ready to ship a final 6.0.2 build of the Microsoft.AspNetCore.Razor.Language, MIcrosoft.CodeAnalysis.Razor etc packages as yet particularly since we'd like to get more changes in to the 6.0.200 branch. My plan was to have these versions follow the same scheme that Roslyn uses for their packages: https://github.com/dotnet/sdk/blob/release/6.0.2xx/eng/Versions.props#L124-L129 and only produce a stable build right around the time 7.0 releases.

dougbu: Yes, this branch now produces stable versions and aligns w/ the (also stable) versions previously flowing into razor-tooling from aspnetcore. Nothing here currently goes to NuGet anyhow but it seemed to make more sense given everything else flowing into dotnet/razor-tooling.

That said, `revert` if you like but you may hit downgrades in later dotnet/razor-tooling PRs unless you also block dotnet/razor-tooling#5801.

PR #12, opened 2021-12-02
Reverting version stabilization changes

pranavkm:
Per https://github.com/dotnet/razor-compiler/pull/10/files#r760754594, we'd like to hold off on producing stable version packages right away. We'd also like to use the versioning scheme that Roslyn uses. Rolling back some parts of https://github.com/dotnet/razor-compiler/pull/10/files

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -4,11 +4,11 @@
     <MajorVersion>6</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>2</PatchVersion>
-    <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>1</PreReleaseVersionLabel>
     <!--
         When StabilizePackageVersion is set to 'true', this branch will produce stable outputs for 'Shipping' packages
     -->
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -4,11 +4,11 @@
     <MajorVersion>6</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>2</PatchVersion>
-    <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>1</PreReleaseVersionLabel>
```
wtgodbe, line 7: Is using a digit here instead of a string intentional? Is that to achieve the roslyn-style versioning?

pranavkm: Yup. Here's the inspiration: https://github.com/dotnet/roslyn/blob/main/eng/Versions.props#L11

wtgodbe (approved):

eng/Versions.props
dougbu: This is so weird.

Note as well that 6.0.1 Razor compiler bits are still flowing out of dotnet/aspnetcore and some of those e.g. https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Razor.Extensions/6.0.0 will be included in every 6.0.x release.

PR #13, opened 2021-12-02
Handle nested self-closing tags

ryanbrandenburg:
- The `TagHelperParseTreeRewriter` does not currently account for self-closing tags when handling nested MarkupElements, which can cause the SyntaxTree to omit the EndTag.
- When the EndTag is omitted the SyntaxTree can have a shorter length than the source document, which causes big problems when a user tries to access anything after the supposed end of the document, and causes "Disco Colors".
- To experience this bug in production open a razor file in the new Razor LSP editor and create a component with the following contents":
```razor
<div @onclick="() => {}"><div/></div>
<strong>Miscolored!</strong>
```
image


NTaylorMullen (approved):
Glorious! IIRC I was actually the one who implemented the start/end tag handling agesssss ago. I'm to blame for the original bug 😄

ryanbrandenburg:
I simulated a compiler insertion to razor-tooling and confirmed this fixes the issue:
image

NTaylorMullen:
Lol @pranavkm I see how it is, the inglorious gray check mark!
image

pranavkm (approved):

PR #14, opened 2021-12-02
Produce transport packages for tooling

pranavkm:


pranavkm:
Here's what the consuming end looks like: https://github.com/dotnet/razor-tooling/compare/prkrishn/use-transport-packages?expand=1

Essentially tooling and SDK will be updated to consume these two new packages, and would no longer consume the product Razor.Language, CodeAnalysis.Razor and Mvc.Razor.Extension packages. In addition, in the 6.0.2xx branch, we'll no longer produce these packages to avoid conflating with the servicing versions (at least until we can figure out how to stop building them in 6.0 @ aspnetcore.

dougbu:
- Doesn't Microsoft.AspNetCore.Razor.Microbenchmarks also ship and cause flip-flopping between this branch and release/6.0 in dotnet/aspnetcore❔
- I'm also unsure what will happen w/ the **non**-shipping packages that the sdk repo consumes for 6.0.2xx

dougbu:
> I'm also unsure what will happen w/ the non-shipping packages that the sdk repo consumes for 6.0.2xx

Actually, I'm pretty sure flip-flopping will occur e.g. with at least Microsoft.AspNetCore.Razor.Internal.SourceGenerator.Transport (which dotnet/aspnetcore probably needs to provide in 6.0.x for the SDK. Suggest marking at least that as non-packable.

pranavkm:
> Doesn't Microsoft.AspNetCore.Razor.Microbenchmarks also ship and cause flip-flopping between this branch and release/6.0 in dotnet/aspnetcore❔

Looks like it's marked as packable and we don't need for it to be. I don't think anything downstream consumes it (it's a benchmarking app so 🤷🏽)

> Actually, I'm pretty sure flip-flopping will occur e.g. with at least Microsoft.AspNetCore.Razor.Internal.SourceGenerator.Transport (which dotnet/aspnetcore probably needs to provide in 6.0.x for the SDK). Suggest marking at least that as non-packable

Yup, so this one we definitely want to stop building out of aspnetcore. It's a purely tooling asset. I can send a PR to do that. 

> I'm also unsure what will happen w/ the non-shipping packages that the sdk repo consumes for 6.0.2xx

My plan was to also update SDK to use these transport packages so it's no longer consuming any of the runtime targeted packages.

dougbu:
> Yup, so this one we definitely want to stop building out of aspnetcore.

How will we update the source generator in the 6.0.1xx (not &hellip;2xx) SDK if you do this❔

pranavkm:
> How will we update the source generator in the 6.0.1xx (not …2xx) SDK if you do this❔

Ooh good point. Ok, I should just rename the package in that case in this repo and so that way we would have a clean delineation.

pranavkm:
Alright, 8th time is the charm. This PR now

a) Stops packaging any of the runtime packages
b) Produces new transport packages to be consumed by tooling and SDK
c) Stops packaging the Microbenchmarks package.

Running pack now produces the following output:

```
Successfully created package 'Debug\NonShipping\Microsoft.AspNetCore.Razor.Internal.Transport.6.0.2-dev.nupkg'.
  Successfully created package 'Debug\NonShipping\Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.6.0.2-dev.nupkg'.
  Successfully created package 'Debug\NonShipping\Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.6.0.2-dev.symbols.nupkg'.
  Successfully created package 'Debug\NonShipping\Microsoft.AspNetCore.Razor.Internal.Transport.6.0.2-dev.symbols.nupkg'.
  Successfully created package 'Debug\NonShipping\Microsoft.AspNetCore.Mvc.Razor.Extensions.Version2_X.6.0.2-dev.nupkg'.
  Successfully created package 'Debug\NonShipping\Microsoft.AspNetCore.Mvc.Razor.Extensions.Version2_X.6.0.2-dev.symbols.nupkg'.
```

dougbu:
Isn't there a conflict w/ Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X and Microsoft.AspNetCore.Mvc.Razor.Extensions.Version2_X still (for the flows into razor-tooling, unless you're going to turn those off)❔ Or does the SDK not consume those❔

Generally think it may be safest to avoid any and all duplicated package names between this repo and dotnet/aspnetcore.

pranavkm:
The SDK does not consume those packages. They were solely meant to be consumed by razor-tooling. I guess I could rename those too.

dougbu:
➕ @wtgodbe in case I'm offline when this is ready for final review.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/pack/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.Internal.Transport.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
dougbu, line 1: nit: Why not just rename the Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X project and package❔ Same thing for the Microsoft.AspNetCore.Mvc.Razor.Extensions.Version2_X project / package.

Just wondering if this is about minimizing differences from 'main'

pranavkm:
I think this is good to go. I'll follow up on the razor-tooling changes as the update PR.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.csproj
```diff
@@ -18,8 +19,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\Microsoft.AspNetCore.Razor.Language\src\Microsoft.AspNetCore.Razor.Language.csproj" />
-    <ProjectReference Include="..\..\Microsoft.CodeAnalysis.Razor\src\Microsoft.CodeAnalysis.Razor.csproj" />
+    <ProjectReference Include="..\..\Microsoft.AspNetCore.Razor.Language\src\Microsoft.AspNetCore.Razor.Language.csproj" PrivateAssets="All" />
+    <ProjectReference Include="..\..\Microsoft.CodeAnalysis.Razor\src\Microsoft.CodeAnalysis.Razor.csproj" PrivateAssets="All" />
```
dougbu, line 23: I'm not understanding the `PrivateAssets="All"` additions and new Microsoft.CodeAnalysis.Razor.csproj references. If these projects don't result in packages, do the changes matter❔

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/pack/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.Internal.Transport.csproj
pranavkm: The binaries are referenced by name in Razor-Tooling which I didn't want to change. I ran in to some trouble trying to fix use RootNamespaces and AssemblyNames to make it work correctly. We can clean it up at some point, but it seems kinda unimportant.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.csproj
pranavkm: I can revert it.

pranavkm:
Good to go?

dougbu (approved):
Looks fine to me. May want someone from razor-tooling or sdk repos to confirm this works for them

PR #22, opened 2021-12-22
Enable in-build signing

dougbu:
- also validate signed packages


dougbu:
Did an internal test build in https://dev.azure.com/dnceng/internal/_build/results?buildId=1523327&view=results but have not confirmed much beyond success of that build and existence of assembly and package signatures.

/fyi @mmitche @NTaylorMullen

pranavkm (approved):

PR #29, opened 2022-01-06
Use nullable context when declaring the class for Razor views

pranavkm:
Currently the Razor compiler suppresses nullable annotations within most generated code. https://github.com/dotnet/aspnetcore/issues/39360 tracks addressing this, but that would need a fair bit of work to get right. In the meantime, https://github.com/dotnet/aspnetcore/issues/37510 is affecting a lot of users attempting to turn on nullability in their apps. 

The value of the `@model` directive appears as part of the class declaration for a Razor View: 

```C#
@model MyModel?

// Results in 
#nullable disable
internal sealed class _Views_Home_Index : RazorPage<MyModel?> // This will produce a warning since it's not within a nullable context
```

The `#nullable disable` precludes the ability of the user to be able to specify a nullable `model`. This change updates the compiler and generates the class declaration within a nullable context, as long as the C# version is recent enough to support nullability. The generated code now looks like:

```C#
@model MyModel?

// Results in 
...
#nullable restore // Use the nullability based on the project's global settings
internal sealed class _Views_Home_Index : RazorPage<MyModel?>
#nullable disable
```

In addition to the class declaration, we also need to generate the properties for `@inject` within a nullable context. This is because all views have an implicit `@inject IHtmlHelper<TModel>` where `TModel` is the model type.


pranavkm:
Ping. Could I get a review on this?

src/Microsoft.AspNetCore.Razor.Language/test/CodeGeneration/DefaultDocumentWriterTest.cs
```diff
@@ -213,17 +213,67 @@ public void WriteDocument_WritesClass()
...
             BaseType = "TestBase",
             Interfaces = new List<string> { "IFoo", "IBar", },
             TypeParameters = new List<TypeParameter>
-                {
-                    new TypeParameter() { ParameterName = "TKey", },
-                    new TypeParameter() { ParameterName = "TValue", },
-                },
+            {
+                new TypeParameter() { ParameterName = "TKey", },
+                new TypeParameter() { ParameterName = "TValue", },
```
TanayParikh, line 224: Should we have a constraint here as well?

TanayParikh (approved):
Don't have the most context in this area, but this looks good to me. Just a question regarding one of the tests.

src/Microsoft.AspNetCore.Razor.Language/test/CodeGeneration/DefaultDocumentWriterTest.cs
pranavkm: Nullability constraints? So within a view, there isn't a way to specify constraints for the model `@typeparam` doesn't work for it, It does work for Blazor components, but enabling nullability for it is a bunch more work - Blazor generates an additional class / method for type inference and we have to make sure things flow everywhere. There were a number of issues with the current code generation and I didn't want to risk introducing bugs in a 6.0 release by trying to tackle those.

PR #31, opened 2022-01-12
Use nullable context when declaring the class for Razor views (#29)

pranavkm:
Backporting https://github.com/dotnet/razor-compiler/pull/29 to 6.0.2xx

Currently the Razor compiler suppresses nullable annotations within most generated code. https://github.com/dotnet/aspnetcore/issues/39360 tracks addressing this, but that would need a fair bit of work to get right. In the meantime, https://github.com/dotnet/aspnetcore/issues/37510 is affecting a lot of users attempting to turn on nullability in their apps. 

The value of the `@model` directive appears as part of the class declaration for a Razor View: 

```C#
@model MyModel?

// Results in 
#nullable disable
internal sealed class _Views_Home_Index : RazorPage<MyModel?> // This will produce a warning since it's not within a nullable context
```

The `#nullable disable` precludes the ability of the user to be able to specify a nullable `model`. This change updates the compiler and generates the class declaration within a nullable context, as long as the C# version is recent enough to support nullability. The generated code now looks like:

```C#
@model MyModel?

// Results in 
...
#nullable restore // Use the nullability based on the project's global settings
internal sealed class _Views_Home_Index : RazorPage<MyModel?>
#nullable disable
```

In addition to the class declaration, we also need to generate the properties for `@inject` within a nullable context. This is because all views have an implicit `@inject IHtmlHelper<TModel>` where `TModel` is the model type.


TanayParikh (approved):
Approving Backport

PR #34, opened 2022-01-20
Applies 'global::' across the board for design time and runtime code generation

javiercn:
Fixes https://github.com/dotnet/aspnetcore/issues/18757


src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentLoweringPass.cs
```diff
@@ -56,15 +58,73 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
                 }
             }
 
-            if (count >= 1)
+            if (count == 1)
             {
-                reference.Replace(RewriteAsComponent(node, node.TagHelpers.First(t => t.IsComponentTagHelper())));
+                reference.Replace(RewriteAsComponent(node, node.TagHelpers.Single(n => n.IsComponentTagHelper())));
+            }else if (count > 1)
```
TanayParikh, line 64:
```suggestion
            } else if (count > 1)
```

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentLoweringPass.cs
```diff
@@ -56,15 +58,73 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
...
+                var tagHelper = node.TagHelpers[i];
+                if (!tagHelper.IsComponentTagHelper())
+                {
+                    continue;
+                }
+
+                for (var j = 0; j < usings.Count; j++)
+                {
+                    var usingNamespace = usings[j].Content;
+                    if(string.Equals(tagHelper.ParsedTypeInfo.Value.Namespace.Value, usingNamespace, StringComparison.Ordinal))
```
TanayParikh, line 95:
```suggestion
                    if (string.Equals(tagHelper.ParsedTypeInfo.Value.Namespace.Value, usingNamespace, StringComparison.Ordinal))
```

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentLoweringPass.cs
```diff
@@ -56,15 +58,73 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
...
+                        }
+                        else
+                        {
+                            matched ??= new();
+                            matched.Add(tagHelper);
+                        }
+                    }
+                }
+            }
+            if (matched != null)
```
TanayParikh, line 109:
```suggestion
            }
            
            if (matched != null)
```

src/Microsoft.AspNetCore.Razor.Language/src/Components/TypeNameHelper.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Razor.Language;
+internal class TypeNameHelper
+{
+    private static readonly string[] PredefinedTypeNames = new[] {"bool",
+        "int",
```
TanayParikh, line 12:
```suggestion
    private static readonly string[] PredefinedTypeNames = new[] {
        "bool",
        "int",
```

Also, should we leave a comment about the fast pass logic below in case a new C# type is added (granted that's pretty unlikely).

TanayParikh (approved):
LGTM, just some nits. Regarding the actual baseline test files, just did some spot checks.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
```diff
@@ -502,9 +504,20 @@ public override void WriteComponent(CodeRenderingContext context, ComponentInter
...
+            }
+            else
+            {
+                if (!node.TagName.Contains("."))
+                {
+                    // The tag is not fully qualified
+                    context.CodeWriter.Write(node.Component.ParsedTypeInfo.Value.Namespace);
+                    context.CodeWriter.Write(".");
+                }
+                context.CodeWriter.Write(node.TagName);
```
javiercn, line 520: This is one of the interesting parts where we need to play with the fact that `TypeName` == `TagName` since TypeName inside ParsedTypeInfo contains the generic parameters and we don't want to parse those.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentLoweringPass.cs
```diff
@@ -56,15 +58,73 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
...
             }
 
-            if (count >= 1)
+            if (count == 1)
             {
-                reference.Replace(RewriteAsComponent(node, node.TagHelpers.First(t => t.IsComponentTagHelper())));
+                reference.Replace(RewriteAsComponent(node, node.TagHelpers.Single(n => n.IsComponentTagHelper())));
+            }else if (count > 1)
+            {
+                var tagHelper = GetTagHelperOrAddDiagnostic(node, usings);
```
javiercn, line 66: We now do the analysis for multiple tag helpers usage here since we are using the fully qualified type for referring to them in the generated code and that prevents errors due to ambiguous usage from being caught. (Previously we would have picked the first TH and relied on the generated code to report the ambiguity)

src/Microsoft.AspNetCore.Razor.Language/src/Components/TypeNameHelper.cs
javiercn: I don't think it's necessary given how unlikely it is, and wouldn't help discovery, since by the time something like that caused a bug, it will be self evident on the investigation

src/Microsoft.AspNetCore.Razor.Language/src/Components/TypeNameHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Razor.Language;
+internal class TypeNameHelper
+{
+    private static readonly string[] PredefinedTypeNames = new[] {
```
pranavkm, line 11: I'm a little vary about rolling our own way to globally qualify things / keeping our own list independent of the compiler. 

Are we not able to rely on the compilers' API to do this e.g. SymbolDisplayMiscellaneousOptions.UseSpecialTypes (https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.symboldisplaymiscellaneousoptions?view=roslyn-dotnet-4.0.1) should already know about special types.

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentDesignTimeCodeGenerationTest/AsyncEventHandler_OnElement_ActionEventArgs_Lambda/TestComponent.codegen.cs
```diff
@@ -33,7 +33,7 @@ public partial class TestComponent : Microsoft.AspNetCore.Components.ComponentBa
         #pragma warning disable 1998
         protected override void BuildRenderTree(Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder __builder)
         {
-            __o = Microsoft.AspNetCore.Components.EventCallback.Factory.Create<Microsoft.AspNetCore.Components.Web.MouseEventArgs>(this, 
+            __o = Microsoft.AspNetCore.Components.EventCallback.Factory.Create<global::Microsoft.AspNetCore.Components.Web.MouseEventArgs>(this,
```
pranavkm, line 36: Microsoft.AspNetCore.Components.EventCallback.Factory should be globally qualified. Same for RenderTreeBuilder, ComponentBase etc

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentRuntimeCodeGenerationTest/BodyAndAttributeChildContent/TestComponent.codegen.cs
```diff
@@ -38,7 +38,7 @@ protected override void BuildRenderTree(Microsoft.AspNetCore.Components.Renderin
 #line default
 #line hidden
 #nullable disable
-            __builder.OpenComponent<Test.MyComponent>(2);
+            __builder.OpenComponent<global::Test.MyComponent>(2);
             __builder.AddAttribute(3, "Header", (Microsoft.AspNetCore.Components.RenderFragment<System.String>)(
```
pranavkm: RenderFragment needs to be qualified

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentRuntimeCodeGenerationTest/ComponentWithConstrainedTypeParameters/UseTestComponent.codegen.cs
```diff
@@ -20,7 +20,7 @@ public partial class UseTestComponent : Microsoft.AspNetCore.Components.Componen
         #pragma warning disable 1998
         protected override void BuildRenderTree(Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder __builder)
         {
-            __Blazor.Test.UseTestComponent.TypeInference.CreateTestComponent_0(__builder, 0, 1, 
+            global::__Blazor.Test.UseTestComponent.TypeInference.CreateTestComponent_0(__builder, 0, 1,
```
pranavkm: In the TypeInference method that we codegen, I think the constraints need to have global qualifiers:

```diff
 public static void CreateTestComponent_0<TItem1, TItem2, TItem3>(global::Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder __builder, int seq, int __seq0, TItem1 __arg0, int __seq1, global::System.Collections.Generic.List<TItem2> __arg1, int __seq2, TItem3 __arg2, int __seq3, global::Microsoft.AspNetCore.Components.RenderFragment<TItem2> __arg3)
-            where TItem1 : Image
+          where TItem1 : global::Foo.Bar.Image
            where TItem2 : ITag
            where TItem3 : Image, new()
```

pranavkm (commented):
Very cool!

src/Microsoft.AspNetCore.Razor.Language/src/Components/TypeNameHelper.cs
javiercn: I wasn't aware of that API, but I'm not sure how it can be used here, since I don't have a symbol?

javiercn: I don't see how this API can be used to determine if the string is a keyword, please provide more details on what you expect, otherwise this feedback is not acctionable.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/BasicComponent_Runtime.codegen.cs
```diff
@@ -9,7 +9,7 @@ namespace __GeneratedComponent
     using System.Linq;
     using System.Threading.Tasks;
     using Microsoft.AspNetCore.Components;
-    public partial class AspNetCore_d3c3d6059615673cb46fc4974164d61eabadb890 : Microsoft.AspNetCore.Components.ComponentBase, IDisposable
+    public partial class AspNetCore_d3c3d6059615673cb46fc4974164d61eabadb890 : global::Microsoft.AspNetCore.Components.ComponentBase, IDisposable
     {
         #pragma warning disable 1998
         protected override void BuildRenderTree(Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder __builder)
```
pranavkm, line 15: This one

src/Microsoft.AspNetCore.Razor.Language/src/Components/TypeNameHelper.cs
pranavkm: Aren't all the type names we originally resolve from compiler symbols? We're not resolving them from runtime type names, so why can't we have the compiler resolve these names for us in the first place? 

Same goes for using the compiler APIs - https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.symboldisplayglobalnamespacestyle?view=roslyn-dotnet-4.0.1 - to produce  globally qualified type names in the first place rather than roll our own thing. For e.g. here's a place where we explicitly chose not to use globally qualified name: https://github.com/dotnet/razor-compiler/blob/52505c754be0bc118a0f335eba10742e36eeb965/src/Microsoft.CodeAnalysis.Razor/src/ComponentTagHelperDescriptorProvider.cs#L16-L19. I'm not sure why that code explicitly avoids using a globally qualified format, but it already resolves the special types. 

I hope these are enough pointers for you to take action on.

javiercn: I don't see how we can leverage this APIs. While it could potentially work for tag helper descriptors, there are plenty more things we need to emit, and as far as I can tell, the only information that we have available comes in the form of a string.

I understand this is not ideal, but I believe it will do the trick and is something we can put in place until we are ready to tackle the refactors needed to accomplish this in a more suitable way.

src/Microsoft.AspNetCore.Razor.Language/src/Components/TypeNameHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        if (typeName == null)
+        {
+            throw new ArgumentNullException(nameof(typeName));
+        }
+
+        return GloballyQualifiedTypeName(new StringSegment(typeName));
+    }
+
+    internal static StringSegment GloballyQualifiedTypeName(StringSegment typeName)
```
SteveSandersonMS, line 41: The naming here might trip us up in the future. Could we make it more explicitly reflect what it does?

For example, it would produce an invalid result if you pass in the otherwise-valid type name `String` (it will return `global::String"`, which isn't a thing).

We could:

 * Rename it to `PrependGlobalToTypeName`
 * Or, change the parameter name to `fullyQualifiedTypeName`

In both cases this would clarify that it's the caller's responsibility to ensure the incoming value isn't something that's legal with respect to certain `using` statements, but rather is always legal.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
```diff
@@ -112,7 +112,7 @@ private void WriteCSharpExpressionInnards(CodeRenderingContext context, CSharpEx
                 if (type != null)
                 {
                     context.CodeWriter.Write("(");
-                    context.CodeWriter.Write(type);
+                    context.CodeWriter.Write(TypeNameHelper.GloballyQualifiedTypeName(type));
```
SteveSandersonMS, line 115: It's tricky to trace back all the code that calls into here, but do we know for sure that `type` is always fully-qualified already?

javiercn: I have a check to No-op if the type is already globally qualified. (might not be pushed)

SteveSandersonMS: Also we could save an allocation by not doing this, and instead doing something like:

```
if (TypeNameHelper.CanSafelyPrependGlobal(type))
{
    context.CodeWriter.Write("global::");
}
context.CodeWriter.Write(type);
```

Not sure whether this is a hot-enough code path to warrant caring about allocations though.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
```diff
@@ -502,9 +504,20 @@ public override void WriteComponent(CodeRenderingContext context, ComponentInter
...
+                context.CodeWriter.Write(node.Component.Name);
+            }
+            else
+            {
+                if (!node.TagName.Contains("."))
+                {
+                    // The tag is not fully qualified
+                    context.CodeWriter.Write(node.Component.ParsedTypeInfo.Value.Namespace);
+                    context.CodeWriter.Write(".");
+                }
```
SteveSandersonMS, line 519: If we ever start supporting partially-qualified tags, e.g.:

```razor
@using Microsoft.AspNetCore.Components

<Web.PageTitle>Counter</Web.PageTitle>
```

... then distinguishing based on the presence of `.` would start being buggy. I know we haven't committed to support that, but it will be hard to find these kinds of issues in the future if we do.

I don't have a good suggestion for how to make this more robust, because I don't know whether `node.Component.ParsedTypeInfo.Value.Namespace` refers to the true type namespace, or just the part of the FQN that precedes whatever is in `node.Component.Name`.

Perhaps the best suggestion I have is to centralize any logic that does things like looking for `.` into some common type name helper so that we'd only have one thing to fix, or could at least look for all the code that uses such a helper.

What do you think?

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
```diff
@@ -621,6 +634,7 @@ private void WriteComponentAttributeInnards(CodeRenderingContext context, Compon
                 if (canTypeCheck)
                 {
                     context.CodeWriter.Write("new ");
+                    context.CodeWriter.Write("global::");
                     context.CodeWriter.Write(node.TypeName);
```
SteveSandersonMS, line 638: I'm assuming we always know that `node.TypeName` is fully-qualified.

My uncertainty stems from a comment above where you wrote `TypeName == TagName`, but I'm guessing you mean `TypeName != TagName`, because other code above has to account for the possibility that `node.TagName` may not be fully-qualified.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
```diff
@@ -656,14 +670,15 @@ private void WriteComponentAttributeInnards(CodeRenderingContext context, Compon
...
+                context.CodeWriter.Write("global::");
                 context.CodeWriter.Write(ComponentsApi.EventCallback.FactoryAccessor);
                 context.CodeWriter.Write(".");
                 context.CodeWriter.Write(ComponentsApi.EventCallbackFactory.CreateMethod);
 
                 if (node.TryParseEventCallbackTypeArgument(out StringSegment argument))
                 {
                     context.CodeWriter.Write("<");
-                    context.CodeWriter.Write(argument);
+                    context.CodeWriter.Write(TypeNameHelper.GloballyQualifiedTypeName(argument));
```
SteveSandersonMS, line 681: As usual, just checking we know that `argument` is always fully-qualified here :)

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
```diff
@@ -936,7 +951,7 @@ protected override void WriteReferenceCaptureInnards(CodeRenderingContext contex
             // just want sufficiently similar code that any unknown-identifier or type
             // errors will be equivalent
             var captureTypeName = node.IsComponentCapture
-                ? node.ComponentCaptureTypeName
+                ? TypeNameHelper.GloballyQualifiedTypeName(node.ComponentCaptureTypeName)
```
SteveSandersonMS, line 954: In this case I tried to trace back the execution and it does look like `ComponentCaptureTypeName` would always be fully qualified, but it's an unfortunate weakness of the Razor Compiler codebase that we don't help ourselves keep track of these invariants.

javiercn: Yep, I noticed this as I was adding cases for spots that I missed. 

I've switched the method to check if the type is already globally qualified and do nothing in that case. 

Otherwise its a nightmare to do this at the call-site.

It's unfortunate that we don't properly flow this information, but unfortunately things like the node.TypeName are used in comparisons through the codebase and changing them there has a broad impact.

Improving this situation is something that I hope we can do as we work to unify the compile/design time builds

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDocumentClassifierPass.cs
```diff
@@ -98,7 +98,7 @@ protected override CodeTarget CreateTarget(RazorCodeDocument codeDocument, Razor
         }
         else
         {
-            @class.BaseType = ComponentsApi.ComponentBase.FullTypeName;
+            @class.BaseType = "global::" + ComponentsApi.ComponentBase.FullTypeName;
```
SteveSandersonMS, line 101: I'm a bit worried by adding the `global::` prefix here and below in this file. It means that any other code needs to somehow know that for these things in particular, the prefix is already present and they can't re-prefix it.

Would it be practical to rename this to `GloballyQualifiedBaseType` (and `GloballyQualifiedTypeName` below) so that we can see which other code uses it, and therefore trace through that they interpret the string in the right way?

Or, more broadly, we could stop using `string` to represent all these different kinds of type name, and instead have some struct that tracks what flavour of type name it is and contains the relevant helper methods like `.ToGloballyQualifiedTypeName()`.

Again, I'm aware that this potentially blows up the scope of the PR and don't want to be inventing problems.

SteveSandersonMS: Another approach would be some principle like "we only ever insert the `global::` prefix at the last possible moment when concatentating to C# source code" (though that wouldn't help us keep track of which type names were FQN and which aren't).

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentRuntimeNodeWriter.cs
```diff
@@ -608,22 +608,23 @@ private void WriteComponentAttributeInnards(CodeRenderingContext context, Compon
...
+                context.CodeWriter.Write("global::");
                 context.CodeWriter.Write(ComponentsApi.EventCallback.FactoryAccessor);
                 context.CodeWriter.Write(".");
                 context.CodeWriter.Write(ComponentsApi.EventCallbackFactory.CreateMethod);
 
                 if (node.TryParseEventCallbackTypeArgument(out StringSegment argument))
                 {
                     context.CodeWriter.Write("<");
-                    context.CodeWriter.Write(argument);
+                    context.CodeWriter.Write(TypeNameHelper.GloballyQualifiedTypeName(argument));
```
SteveSandersonMS, line 627: I don't know which one of these is the hot-path case that we use for normal components, but whichever one that is, would it be worth avoiding the allocation done by `GloballyQualifiedTypeName`, and instead having explicit logic like:

```razor
if (shouldAddPrefix)
{
    Write("global::");
}
Write(typeName);
```

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentLoweringPass.cs
SteveSandersonMS: Ah, this relates to a question I asked.

To avoid this knowledge getting lost, could you put that info into the code somehow? Also should the previous logic that adds a diagnostic on ambiguity be removed now?

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
SteveSandersonMS: > I have a check to No-op if the type is already globally qualified

I didn't mean globally qualified; I meant fully qualified. It's valid to prefix `global::` to `System.String`, but not valid to prefix it to `String`, for example.

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentDesignTimeCodeGenerationTest/AsyncEventHandler_OnElement_ActionEventArgs_Lambda/TestComponent.codegen.cs
javiercn: Done.

I've used a regex over the generated code baselines `(?<!global::)Microsoft.AspNetCore.Components` to find all the places where we didn't put global:: until I couldn't find any matches.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
javiercn: Yep, all types that get here are fully qualified

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
javiercn: I've updated this logic to use a helper that checks if the type is already fully qualified

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDocumentClassifierPass.cs
javiercn: I've already looked into what other code uses this, and it's read pretty much just to emit the actual code, so I'm not worried about it. For example, this doesn't change how we discover components because we compile the dll and then work of off that.

src/Microsoft.AspNetCore.Razor.Language/src/Components/TypeNameHelper.cs
javiercn: I change this to `WriteGloballyQualifiedName` as well as removed the allocation

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
javiercn: removed the allocation

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentLoweringPass.cs
```diff
@@ -56,15 +58,74 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
...
+
+                reference.Replace(RewriteAsComponent(node, tagHelper));
             }
             else
             {
                 reference.Replace(RewriteAsElement(node));
             }
         }
+
+        static TagHelperDescriptor GetTagHelperOrAddDiagnostic(TagHelperIntermediateNode node, IReadOnlyList<UsingDirectiveIntermediateNode> usings)
```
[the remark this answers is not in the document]
javiercn: Added a comment

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentRuntimeCodeGenerationTest/ComponentWithConstrainedTypeParameters/UseTestComponent.codegen.cs
javiercn: I'm going to leave this one out, since it's a bit more involved to do as it requires parsing the constraint list and it's not that common. I'll file an issue to follow up.

javiercn:
🆙 📅 Any other feeback?

SteveSandersonMS (approved):
Nice! Updates look good, and thanks for checking on the various questions.

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentRuntimeCodeGenerationTest/ComponentWithConstrainedTypeParameters/UseTestComponent.codegen.cs
javiercn: On second thought, we actually shouldn't do this, and the reason is that we "pipe through" these constraints from the directive, so it's code that the user writes and if there's an ambiguity there, they can fix it themselves.

We only care about the places where we generate code that doesn't directly come from user input (events, components, etc.) where the user doesn't have any way control over what gets generated.

javiercn: Note that in your example, we can't know if the type is fully or partially qualified.

SteveSandersonMS:
Just a side-conversation on work on the Razor compiler: does anyone have any good suggestions for how we can make the core review workflow reasonable when "update baselines" generates 1000+ changes? The specific problems I encountered were:

* Couldn't use GitHub's "Files" tab to view any changes in the browser, as the sheer amount of changes cased the web browser to hang. Instead I had to use the GitHub VS Code extension to review this.
* Due to limitations in GitHub VS Code extension, couldn't Ctrl+F find any review comments. There were a couple of conversations going on in here that I wanted to read, but didn't manage to find them manually among the large slew of conversations that are marked as "resolved".

I know we plan to do a lot more Razor Compiler work soon, so would be great to come up with any system to improve this. For example:

 * We could choose not to push any commits to GitHub that update the baselines until review is done. However I know that's inconvenient for the person doing the PR, plus sometimes in code review you actually want to see some baseline changes
 * We could choose not to mark any conversations as "resolved" until it's all completely signed off, or possibly not resolve them at all, so it remains possible to ctrl+F search through them.

Any other ideas @javiercn? How did you review this, @pranavkm?

javiercn:
@SteveSandersonMS this is a particularly challenging one because we were changing the code generation part in a very cross-cutting way. I don't expect most changes to be this impactful and to be more localized in general.

In general, when I had to deal with reviews like this, as an author you probably saw that I updated code and baselines in separate commits as to make it easier to see on each individual commit; in the same vein, as a reviewer, if the PR has these sort of changes and makes it hard to review on GH, I pull it down and perform a `git diff -- <<path>>` to review things and narrow down the changes.

When I started working on this I was planning on breaking down the PR using a "case by case" approach, but that caused a lot of extra work as its easy to make multiple changes or discover that you need to make changes because they don't play well with other required changes (they are not independent).

In general, even if you need to go through lots of files, changes in these files are very similar.

There's obviously room for improvement here, but in general it's hard to do this other than breaking it up in individual commits/updates (some of which will still update hundreds of baseline files)

In my own testing, what I did was craft a regex and use VS Code search to find places where I missed using `global::` until I got all the places we care about.

Its also hard to predict that we didn't break anything, other than by the fact that we have a myriad of tests and that they all check that the code compiles successfully. Which means that if we mess something up, its not common nor critical (hopefully), but we can never be sure.

PR #36, opened 2022-01-20
Stop Eating Optional `;` after `@typeparam`

TanayParikh:
We'd [implemented](https://github.com/dotnet/aspnetcore/pull/37148) optional `;` for `@typeparam` however by eating the token we were causing [issues with the syntax tree](https://github.com/dotnet/razor-tooling/issues/4439#issuecomment-1013555833). @ryanbrandenburg provided [a failing test](https://github.com/dotnet/razor-compiler/commit/f4568a0c38f8883b27033b578def10c225f3f389) so implemented this fix accordingly.

Fixes: https://github.com/dotnet/aspnetcore/issues/39616


src/Microsoft.AspNetCore.Razor.Language/src/Legacy/CSharpCodeParser.cs
```diff
@@ -1490,8 +1490,6 @@ private void ParseExtensibleDirective(in SyntaxListBuilder<RazorSyntaxNode> buil
                                     {
                                         if (At(SyntaxKind.Semicolon))
                                         {
-                                            // Consume the ending ';'
-                                            EatCurrentToken();
```
pranavkm, line 1494: Do we know what it means to not consume the ending `;`?

TanayParikh: From my understanding the `;` ends up being subsequently consumed here:

https://github.com/dotnet/razor-compiler/blob/3e481fe2cfd552ea9a88d3f73898965100a9af24/src/Microsoft.AspNetCore.Razor.Language/src/Legacy/CSharpCodeParser.cs#L1536-L1563

We're just reverting some ([recent~ish](https://github.com/dotnet/aspnetcore/pull/37148/files#diff-949d4c264e4813fb5d32fa77551f1288738faf16b84df43239c88172cadd83f9R1494)) changes partially.

pranavkm (approved):

ryanbrandenburg (approved):
I still don't really understand the cspans files or why all those .mappings.txt's changed, but it seems the the ";" is now included in .stree.txt and the verifier doesn't fail so we know the SyntaxTree length should be correct, so I'm happy.

PR #44, opened 2022-01-24
Fixes creating constraint string from type parameter symbol

javiercn:
Fixes https://github.com/dotnet/aspnetcore/issues/38479 https://github.com/dotnet/aspnetcore/issues/25588


src/Microsoft.CodeAnalysis.Razor/src/ComponentTagHelperDescriptorProvider.cs
```diff
@@ -270,36 +270,35 @@ private void CreateTypeParameterProperty(TagHelperDescriptorBuilder builder, ITy
...
             if (typeParameter.HasUnmanagedTypeConstraint)
             {
-                constraintString.Append(", unmanaged");
+                list.Add("unmanaged");
             }
             if (typeParameter.HasValueTypeConstraint)
             {
-                constraintString.Append(", struct");
+                list.Add("struct");
             }
```
javiercn, line 297: The issue was that each block after the one above assumed that there was already a constraint and pre-pended a comma.

Instead of doing that, we collect the constraints on a list and then join them together.

SteveSandersonMS: Makes sense. Good fix.

In the new baselines in this PR, I don't see any cases where constraints are joined with commas. Do we definitely have coverage of this?

src/Microsoft.CodeAnalysis.Razor/src/ComponentTagHelperDescriptorProvider.cs
```diff
@@ -270,36 +270,35 @@ private void CreateTypeParameterProperty(TagHelperDescriptorBuilder builder, ITy
             // things like constructor constraints and not null constraints in the
             // type parameter so we create a single string representation of all the constraints
             // here.
-            var constraintString = new StringBuilder();
-            if (typeParameter.ConstraintTypes.Any())
+            var list = new List<string>();
```
pranavkm, line 273: We could remove an allocation by lazily initializing this, assuming this code path is always called and not only if the parameter has some constraint. An alternative would to be add a `AppendToStringBuilder` that conditionally adds the commas to avoid the interim string allocation from `string.Join` that happens later on in this method.

pranavkm (approved):

src/Microsoft.CodeAnalysis.Razor/src/ComponentTagHelperDescriptorProvider.cs
javiercn: Not going to remove the allocation unless we have prove that it matters in a benchmark. This code used to work and we broke it for the sake of removing the allocation without a clear benefit. Let's keep the code simple and understandable (and hopefully bug free)

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentDesignTimeCodeGenerationTest/ComponentWithConstrainedTypeParameters/UseTestComponent.codegen.cs
```diff
@@ -89,9 +89,9 @@ namespace __Blazor.Test.UseTestComponent
     internal static class TypeInference
     {
         public static void CreateTestComponent_0<TItem1, TItem2, TItem3>(global::Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder __builder, int seq, int __seq0, TItem1 __arg0, int __seq1, global::System.Collections.Generic.List<TItem2> __arg1, int __seq2, TItem3 __arg2, int __seq3, global::Microsoft.AspNetCore.Components.RenderFragment<TItem2> __arg3)
-            where TItem1 : Image
-            where TItem2 : ITag
-            where TItem3 : Image, new()
+            where TItem1 : global::Image
```
pranavkm, line 92: This doesn't look correct since it's not a FQN.

javiercn:
🆙📅 this now includes an additional fix that we missed (since the fix is contained on the same method and we get it for free)

We were not properly generating the type names in the where clause, and this manifested on type and interface constraints that were not being qualified appropriately as well as on self referencing constraints that weren't also being properly generated either.

We switched to use the logic from roslyn to generate the display name for fully qualified type name of the constraint. In addition to that, we reordered the constraints to make sure that they appear in the right order.

This fixes first the issue with the generated constraint when there was no base type, as well as the issue with specifying a base type or interface constraints

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentDesignTimeCodeGenerationTest/ComponentWithConstrainedTypeParameters/UseTestComponent.codegen.cs
javiercn:
```
        var classes = @"
public class Image
{
    public string url { get; set; }
    public int id { get; set; }

    public Image()
    {
        url = ""https://example.com/default.png"";
        id = 1;
    }
}

public interface ITag
{
    string description { get; set; }
}

public class Tag : ITag
{
    public string description { get; set; }
}
";

        AdditionalSyntaxTrees.Add(Parse(classes));
```

javiercn: Keep in mind that this data comes from the resolved symbol, so it can't produce the wrong output (otherwise it would be a bug in Roslyn)

PR #45, opened 2022-01-25
Use Roslyn to parse type names

javiercn:
Contributes to https://github.com/dotnet/aspnetcore/issues/38879

The exception source seems to be coming from a place where we split a full type name into its type name and namespace. Instead of using our own "dubious" parsing logic, rely on the roslyn APIs to do the parsing.

This entailed adding a new method to `TypeNameFeature` and flowing that to the type that performs this operation.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/test/DefaultRazorTagHelperBinderPhaseTest.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/test/DefaultRazorTagHelperBinderPhaseTest.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using Microsoft.AspNetCore.Razor.Language.Components;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
 using Microsoft.AspNetCore.Razor.Language.Syntax;
+using Microsoft.CodeAnalysis.Razor;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Razor.Language;
@@ -1081,7 +1082,7 @@ public class DefaultRazorTagHelperBinderPhaseTest : RazorProjectEngineTestBase
             };
         var sourceDocument = CreateComponentTestSourceDocument(@"<Counter />", "C:\\SomeFolder\\SomeProject\\Counter.cshtml");
         var tree = RazorSyntaxTree.Parse(sourceDocument);
-        var visitor = new DefaultRazorTagHelperBinderPhase.ComponentDirectiveVisitor(sourceDocument.FilePath, descriptors, currentNamespace);
+        var visitor = new DefaultRazorTagHelperBinderPhase.ComponentDirectiveVisitor(sourceDocument.FilePath, descriptors, currentNamespace, new DefaultTypeNameFeature());
 
         // Act
         visitor.Visit(tree);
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.Razor.Language/test/DefaultRazorTagHelperBinderPhaseTest.cs
```diff
@@ -1346,12 +1347,12 @@ public void IsTagHelperFromMangledClass_WorksAsExpected()
...
         {
                 { "", false, "", ""},
-                { ".", true, "", ""},
+                { ".", true, "", "."},
                 { "Foo", true, "", "Foo"},
                 { "SomeProject.Foo", true, "SomeProject", "Foo"},
                 { "SomeProject.Foo<Bar>", true, "SomeProject", "Foo<Bar>"},
                 { "SomeProject.Foo<Bar.Baz>", true, "SomeProject", "Foo<Bar.Baz>"},
-                { "SomeProject.Foo<Bar.Baz>>", true, "", "SomeProject.Foo<Bar.Baz>>"},
+                { "SomeProject.Foo<Bar.Baz>>", true, "SomeProject", "Foo<Bar.Baz>>"},
```
SteveSandersonMS, line 1355: This is weird. Is it testing for a syntax error? Do we care about what the parsing output would be in that case?

src/Microsoft.CodeAnalysis.Razor/src/DefaultTypeNameFeature.cs
```diff
@@ -66,4 +66,37 @@ public override bool IsLambda(string expression)
...
+        typeNamespace = null;
+        typeName = null;
+        try
+        {
+            var parsed = SyntaxFactory.ParseTypeName(type);
+            switch (parsed)
+            {
+                case GenericNameSyntax genericNameSyntax:
+                    typeName = genericNameSyntax.ToFullString();
+                    typeNamespace = "";
```
SteveSandersonMS, line 81: When does this happen? Getting an empty namespace is pretty strange. But looking at the test cases, I didn't spot one where the namespace actually came back as empty (at least, when `type` was fully qualified).

javiercn: IEnumerable<Hello>

src/Microsoft.AspNetCore.Razor.Language/test/DefaultRazorTagHelperBinderPhaseTest.cs
javiercn: I don't think we care about these cases. I'm not sure they represent "real" scenarios and the logic for parsing is "dubious" at the very least, so now we at least we don't do our own parsing.

src/Microsoft.CodeAnalysis.Razor/src/DefaultTypeNameFeature.cs
javiercn: I might add some additional test cases, I didn't want to invest more time on it before validating the approach E2E with the tests. I also need to check with tooling folks if there's any drawback WRT to tooling scenarios (which I don't think there is)

SteveSandersonMS: I edited my comment before I saw your reply, but possibly after you posted your reply.

> at least, when type was fully qualified

OK, so I guess it's more that "namespace is only given when the syntax is `QualifiedNameSyntax`" in which case cool.

src/Microsoft.AspNetCore.Razor.Language/test/DefaultRazorTagHelperBinderPhaseTest.cs
SteveSandersonMS: OK, fair enough

javiercn:
There's a follow-up PR to get rid of more "in-house" type parsing in favor of leveraging  roslyn APIs following the same approach employed here.

SteveSandersonMS (approved):
Looks good!

pranavkm (approved):
Nice! My only concern is that we'd updated the previous implementation to use StringSegments since this API featured at the top of allocations which this would now undo.

pranavkm:
@javiercn can we merge this?

TanayParikh (approved):
:shipit:

kdubau (approved):

PR #50, opened 2022-01-25
[release/6.0.2xx] Remove invalid post-build parameter

dougbu:
- template does not support a `pool` parameter


dougbu:
Looks like the recent Arcade update broke this branch too. The parameter was previously ignored but is now causing errors. Checking other branches&hellip;

TanayParikh (approved):

PR #52, opened 2022-01-25
[main] Remove invalid post-build parameter

dougbu:
- template does not support a `pool` parameter


TanayParikh (approved):

dougbu:
That's it for our reaction to latest Arcade update

PR #60, opened 2022-01-31
Unify comparer implementations and improve performance

sharwell:
* Fix cases where `GetHashCode` can return different values for two instances where `Equals` would return true
* Avoid expensive `OrderBy` operations inside of `Equals`

⚠️ This pull request contains _behavior_ changes that have not been validated inside the product. While the new implementations improve consistency between `GetHashCode` and `Equals` and reduce allocation overhead for issues like AB#1464834, it simply assumes that these types are constructed in ways that naturally lead to equivalent ordering in the underlying collections.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
@@ -1,7 +1,5 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
-
-#nullable disable
 
 using System;
 using System.Collections.Generic;
@@ -9,7 +7,7 @@ using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Razor.Language;
 
-internal class AllowedChildTagDescriptorComparer : IEqualityComparer<AllowedChildTagDescriptor>
+internal sealed class AllowedChildTagDescriptorComparer : IEqualityComparer<AllowedChildTagDescriptor?>
 {
     /// <summary>
     /// A default instance of the <see cref="AllowedChildTagDescriptorComparer"/>.
@@ -22,16 +20,20 @@ internal class AllowedChildTagDescriptorComparer : IEqualityComparer<AllowedChil
     }
 
     /// <inheritdoc />
-    public virtual bool Equals(
-        AllowedChildTagDescriptor descriptorX,
-        AllowedChildTagDescriptor descriptorY)
+    public bool Equals(
+        AllowedChildTagDescriptor? descriptorX,
+        AllowedChildTagDescriptor? descriptorY)
     {
         if (object.ReferenceEquals(descriptorX, descriptorY))
         {
             return true;
         }
 
-        if (descriptorX == null ^ descriptorY == null)
+        if (descriptorX is null)
+        {
+            return descriptorY is null;
+        }
+        else if (descriptorY is null)
         {
             return false;
         }
@@ -42,8 +44,13 @@ internal class AllowedChildTagDescriptorComparer : IEqualityComparer<AllowedChil
     }
 
     /// <inheritdoc />
-    public virtual int GetHashCode(AllowedChildTagDescriptor descriptor)
+    public int GetHashCode(AllowedChildTagDescriptor? descriptor)
     {
+        if (descriptor is null)
+        {
+            return 0;
+        }
+
         var hash = HashCodeCombiner.Start();
         hash.Add(descriptor.Name, StringComparer.Ordinal);
 
--- a/src/Microsoft.AspNetCore.Razor.Language/src/BoundAttributeDescriptorComparer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/src/BoundAttributeDescriptorComparer.cs
@@ -1,16 +1,13 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System;
 using System.Collections.Generic;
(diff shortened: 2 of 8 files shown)


sharwell:
Will resubmit this as 6 separate pull requests to ensure the history captures each distinct commit.

src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
```diff
@@ -1,15 +1,13 @@
...
-#nullable disable
-
 using System;
 using System.Collections.Generic;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Razor.Language;
 
-internal class AllowedChildTagDescriptorComparer : IEqualityComparer<AllowedChildTagDescriptor>
+internal sealed class AllowedChildTagDescriptorComparer : IEqualityComparer<AllowedChildTagDescriptor?>
```
sharwell, line 10: 📝 All `sealed` additions are in classes where the constructor was already `private`.

src/Microsoft.AspNetCore.Razor.Language/src/BoundAttributeDescriptorComparer.cs
```diff
@@ -46,17 +43,14 @@ public bool Equals(BoundAttributeDescriptor descriptorX, BoundAttributeDescripto
             string.Equals(descriptorX.IndexerTypeName, descriptorY.IndexerTypeName, StringComparison.Ordinal) &&
             string.Equals(descriptorX.Documentation, descriptorY.Documentation, StringComparison.Ordinal) &&
             string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal) &&
-            Enumerable.SequenceEqual(
-                descriptorX.Metadata.OrderBy(propertyX => propertyX.Key, StringComparer.Ordinal),
-                descriptorY.Metadata.OrderBy(propertyY => propertyY.Key, StringComparer.Ordinal));
+            ComparerUtilities.Equals(descriptorX.BoundAttributeParameters, descriptorY.BoundAttributeParameters, BoundAttributeParameterDescriptorComparer.Default) &&
```
sharwell, line 46: 📝 Comparison of `BoundAttributeParameters` is new in `Equals`. It was already part of `GetHashCode`.

src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
```diff
@@ -22,28 +20,31 @@ private AllowedChildTagDescriptorComparer()
...
     {
         if (object.ReferenceEquals(descriptorX, descriptorY))
         {
             return true;
         }
 
-        if (descriptorX == null ^ descriptorY == null)
-        {
+        if (descriptorX is null)
+            return descriptorY is null;
```
pranavkm, line 33:
```suggestion
        if (descriptorX is null)
        {
            return descriptorY is null;
        }
```

src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
```diff
@@ -22,28 +20,31 @@ private AllowedChildTagDescriptorComparer()
...
+        else if (descriptorY is null)
             return false;
```
pranavkm, line 35:
```suggestion
        else if (descriptorY is null)
        {
            return false;
        }
```

src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
```diff
@@ -22,28 +20,31 @@ private AllowedChildTagDescriptorComparer()
...
             string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal);
     }
 
     /// <inheritdoc />
-    public virtual int GetHashCode(AllowedChildTagDescriptor descriptor)
+    public int GetHashCode(AllowedChildTagDescriptor? descriptor)
     {
+        if (descriptor is null)
+            return 0;
+
```
pranavkm, line 47:
```suggestion
        if (descriptor is null)
        {
            return 0;
        }

```

src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
sharwell: This style isn't configured from this repository, and deviates from the one described in dotnet/runtime. I'll make sure that I don't remove any braces already present, but will leave the others for a follow-up if/when analyzer enforcement for this style is enabled for the repository.

sharwell: ➡️ All braces which were part of the original code and weren't rewritten as part of this pull request are now restored.

sharwell: ➡️ Rebased and edited commits to account for code style analyzers that are now enabled

src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
sharwell: ➡️ Rebased and edited commits to account for code style analyzers that are now enabled

src/Microsoft.AspNetCore.Razor.Language/src/AllowedChildTagDescriptorComparer.cs
sharwell: ➡️ Rebased and edited commits to account for code style analyzers that are now enabled

pranavkm (approved):
Thanks!

FYI @javiercn

PR #62, opened 2022-02-01
Enforce IDE0011 (Add braces) on build

sharwell:
This style follows the configuration from dotnet/aspnetcore, which intentionally deviates from the recommendation made by dotnet/runtime and dotnet/roslyn.

My recommendation is to alter this pull request to adopt the rules from dotnet/runtime instead of the more restricted rules from dotnet/aspnetcore. In the absence of that, my fallback recommendation is to use a merge commit here instead of squash merge.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -52,6 +52,9 @@ csharp_new_line_before_members_in_anonymous_types = true
 # Namespace settings
 csharp_style_namespace_declarations = file_scoped
 
+# Prefer curly braces even for one line of code
+csharp_prefer_braces = true
+
 [*.{xml,config,*proj,nuspec,props,resx,targets,yml,tasks}]
 indent_size = 2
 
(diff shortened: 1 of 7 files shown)


.editorconfig
```diff
@@ -52,6 +52,9 @@ csharp_new_line_before_members_in_anonymous_types = true
 # Namespace settings
 csharp_style_namespace_declarations = file_scoped
 
+# Prefer curly braces even for one line of code
+csharp_prefer_braces = true
```
sharwell, line 56: 📝 The recommended style for this feature is `when_multiline`:

```suggestion
csharp_prefer_braces = when_multiline
```

pranavkm: ^ @Pilchie

pranavkm (approved):
Thanks!

.editorconfig
Pilchie: OTOH, no single line embedded statements should be allowed either, so...

PR #69, opened 2022-02-03
Avoid globally qualifying user defined types

javiercn:
Fixes https://github.com/dotnet/aspnetcore/issues/39950


src/Microsoft.CodeAnalysis.Razor/src/ComponentTagHelperDescriptorProvider.cs
```diff
@@ -181,7 +181,7 @@ private void CreateProperty(TagHelperDescriptorBuilder builder, IPropertySymbol
             pb.TypeName = property.Type.ToDisplayString(FullNameTypeDisplayFormat);
             pb.SetPropertyName(property.Name);
             pb.IsEditorRequired = property.GetAttributes().Any(a => a.AttributeClass.ToDisplayString() == "Microsoft.AspNetCore.Components.EditorRequiredAttribute");
-
+            pb.SetGloballyQualifiedTypeName(property.Type.ToDisplayString(GloballyQualifiedFullNameTypeDisplayFormat));
```
javiercn, line 184: The key is here, where we capture the globally qualified type name for the type. This comes from the symbol, and we re-use it later on.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentNodeWriter.cs
```diff
@@ -294,7 +294,7 @@ protected List<TypeInferenceMethodParameter> GetTypeInferenceMethodParameters(Co
                     typeName = attribute.TypeName;
                     if (attribute.BoundAttribute != null && !attribute.BoundAttribute.IsGenericTypedProperty())
                     {
-                        typeName = "global::" + typeName;
+                        typeName = typeName.StartsWith("global::",StringComparison.Ordinal) ? typeName : $"global::{typeName}";
```
pranavkm, line 297:
```suggestion
                        typeName = typeName.StartsWith("global::", StringComparison.Ordinal) ? typeName : $"global::{typeName}";
```

pranavkm (approved):

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentGenericTypePass.cs
```diff
@@ -350,6 +350,18 @@ private void RewriteTypeNames(TypeNameRewriter rewriter, ComponentIntermediateNo
...
+                    // Given that TItem can have been defined explicitly by the user to a partially
+                    // qualified type, (like MyType), we check if the globally qualified type name
+                    // contains "global::" which will be the case in all cases as we've computed
+                    // this information from the Roslyn symbol except for when the symbol is a generic
+                    // type parameter. In which case, we mark it with an additional annotation to
+                    // acount for that during code generation and avoid trying to fully qualify
+                    // the type name.
+                    if (!globallyQualifiedTypeName.StartsWith("global::", StringComparison.Ordinal))
+                    {
+                        attribute.Annotations.Add(ComponentMetadata.Component.ExplicitTypeNameKey, true);
```
javiercn, line 363: We use this metadata to signal that we replaced something like TItem with potential user input, like MyType (which we can't qualify) to avoid prepending `global::` incorrectly.

src/Microsoft.CodeAnalysis.Razor/src/GlobalQualifiedTypeNameRewriter.cs
```diff
@@ -59,7 +59,21 @@ public override SyntaxNode VisitQualifiedName(QualifiedNameSyntax node)
             node = (QualifiedNameSyntax)base.VisitQualifiedName(node);
 
             // Rewriting these is complicated, best to just tostring and parse again.
-            return SyntaxFactory.ParseTypeName("global::" + node.ToString());
+            return SyntaxFactory.ParseTypeName(IsGloballyQualified(node) ? node.ToString() : "global::" + node.ToString());
```
TanayParikh, line 62: Nit;

```suggestion
            return SyntaxFactory.ParseTypeName((IsGloballyQualified(node) ? node.ToString() : "global::") + node.ToString());
```

TanayParikh (approved):

src/Microsoft.CodeAnalysis.Razor/src/GlobalQualifiedTypeNameRewriter.cs
javiercn: That parenthesis is unnecessary, `+` has precendence over `?:` https://docs.microsoft.com/en-us/cpp/c-language/precedence-and-order-of-evaluation?view=msvc-170

PR #70, opened 2022-02-11
Support @bind:get, @bind:set, @bind:after

javiercn:
This PR adds support for this new form of binding.

There are still details to be figured out around error handling and reporting as well as some additional tests involving generics, but the core support for the primitives is implemented for elements and components.

## Disclaimer ## the way the PR/Commits are structured will make our reviewing life's miserable, so we need to discuss how best break down this PR for easier review


src/Microsoft.CodeAnalysis.Razor/src/BindTagHelperDescriptorProvider.cs
```diff
@@ -196,6 +197,35 @@ private TagHelperDescriptor CreateFallbackBindTagHelper()
 
                 parameter.SetPropertyName("Culture");
             });
+
+            attribute.BindAttributeParameter(parameter =>
+            {
+                parameter.Name = "get";
+                parameter.TypeName = typeof(Delegate).FullName;
```
SteveSandersonMS, line 204: I'm a bit surprised this is marked as a delegate, since in the user's sources it looks like an arbitrary expression of any type, e.g.:

`<elem @bind:get="@personName">`

However I can see from your test cases that this already works. Or at least the compiler output is correct. Is it possible that we need to type this as `object` in order to get correct editor behavior, or am I missing how this works?

SteveSandersonMS: Same in the other copy of this code below too.

src/Microsoft.CodeAnalysis.Razor/test/BindTagHelperDescriptorProviderTest.cs
```diff
@@ -374,93 +436,134 @@ public class BindAttributes
...
+
+            // Invariants
+            Assert.Empty(parameter.Diagnostics);
+            Assert.False(parameter.HasErrors);
+            Assert.Equal(ComponentMetadata.Bind.TagHelperKind, parameter.Kind);
+            Assert.False(parameter.IsDefaultKind());
+
+            Assert.Equal(
+                "Specifies the culture to use for conversions.",
+                parameter.Documentation);
```
SteveSandersonMS, line 555: Should there also be some equivalent parameter documentation for the new directive attributes?

src/Microsoft.AspNetCore.Razor.Language/src/TagMatchingRuleDescriptor.cs
```diff
@@ -65,4 +67,29 @@ public override int GetHashCode()
...
+        var tagName = TagName ?? "*";
+        tagName += TagStructure == TagStructure.WithoutEndTag ? "/" : "";
+        return $"{TagName ?? "*"}[{string.Join(", ", Attributes.Select(a => DescribeAttribute(a)))}]";
+        static string DescribeAttribute(RequiredAttributeDescriptor attribute)
+        {
+            var name = attribute.Name switch
+            {
+                null => "*",
+                var prefix when attribute.NameComparison == RequiredAttributeDescriptor.NameComparisonMode.PrefixMatch => $"^{prefix}",
+                var full => $"{full}",
```
SteveSandersonMS, line 82:
```suggestion
                var full => full,
```

SteveSandersonMS: Hmm, that was meant to be posted as a suggestion, but I'm using the VS Code GitHub plugin for this review (to avoid all the baselines) and am not sure if it handles suggestions like the web interface does.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -416,18 +491,37 @@ private IntermediateNode[] RewriteUsage(IntermediateNode parent, BindEntry bindE
...
+                        {
+                            [ComponentMetadata.Common.OriginalAttributeName] = getNode.OriginalAttributeName,
+                        },
+                    AttributeName = valueAttributeName,
+                    BoundAttribute = valueAttribute, // Might be null if it doesn't match a component attribute
+                    PropertyName = valueAttribute?.GetPropertyName(),
+                    TagHelper = valueAttribute == null ? null : getNode.TagHelper,
+                    TypeName = valueAttribute?.IsWeaklyTyped() == false ? valueAttribute.TypeName : null,
+                };
+            }
```
SteveSandersonMS, line 524: These two blocks of code look like maybe the same thing, except one takes `node` and the other takes `getNode`. Could they be coalesced into a single block acting upon some new `var nodeToUse = node ?? getNode`?

Perhaps there's some subtle difference I've missed.

SteveSandersonMS: Same with various other seemingly-duplicated blocks below.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -491,7 +621,7 @@ private IntermediateNode[] RewriteUsage(IntermediateNode parent, BindEntry bindE
 
     private bool TryParseBindAttribute(BindEntry bindEntry, out string valueAttributeName)
     {
-        var attributeName = bindEntry.BindNode.AttributeName;
+        var attributeName = bindEntry.BindNode?.AttributeName ?? bindEntry.BindGetNode?.AttributeNameWithoutParameter;
```
SteveSandersonMS, line 624: In the future it's going to be super challenging for anyone reading this code to reason about why in one case we're dealing with `AttributeName` and in another case `AttributeNameWithoutParameter`. Is there any way of capturing some of the reasoning? A comment would do, but if there are ways of renaming any of these symbols to make their unique roles more apparent that would be even better.

I think the whole Razor compiler codebase really suffers from being write-only code. It's still workable because of the baseline suite and the fact that we can run it under a debugger. But for the longer term it would be great to try making more of the individual units of code possible to understand in isolation.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -543,14 +674,14 @@ private bool TryParseBindAttribute(BindEntry bindEntry, out string valueAttribut
             return false;
         }
 
-        valueAttributeName = node.TagHelper.GetValueAttributeName() ?? valueAttributeName;
+        valueAttributeName = node?.TagHelper.GetValueAttributeName() ?? getNode?.TagHelper.GetValueAttributeName() ?? valueAttributeName;
```
SteveSandersonMS, line 677: Does it work to do things like this here?

`(node ?? getNode)?.TagHelper.GetValueAttributeName()`

Would it even work to put this coalescing logic onto the `BindEntry` type so as to centralise all this? For example if `BindEntry` had a property `EffectiveBindGetNode => BindNode ?? BindGetNode` then other code consuming it wouldn't need any conditionals, and then we'd only have to put a comment in one place (on `BindEntry`) saying why there is a concept of switching between these two nodes.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -647,19 +781,72 @@ static CSharpExpressionIntermediateNode ExtractEventNodeExpression(TagHelperDire
...
+        // With an asynchronous 'after' method we will generate code as follows:
+        // (TargetAttributeType)(__value => <code> = __value; return RuntimeHelpers.InferAsynchronousDelegate(after)(); }
+
         // Now rewrite the content of the change-handler node. Since it's a component attribute,
         // we don't use the 'BindMethods' wrapper. We expect component attributes to always 'match' on type.
         //
         // __value => <code> = __value
-        changeExpressionTokens.Add(new IntermediateToken()
+
+        switch (setter, after, awaitable)
```
SteveSandersonMS, line 802: Since we're not extracting any values from the patterns being matched, to be honest I find it less readable than a series of if/elseif/else checks. It's exactly equivalent to if/elseif/else except you have to keep scanning back to L802 to remember which each entry in the pattern tuple is which variable.

Perhaps this set of patterns grew over time during implementation or something. Not a big deal but especially given how `awaitable` is only used in two of the cases, and the first and last case are treated the same, it altogether seems like a more convenient representation might be possible now we have the end result.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -668,9 +855,55 @@ static CSharpExpressionIntermediateNode ExtractEventNodeExpression(TagHelperDire
...
         {
-            Content = $"{ComponentsApi.RuntimeHelpers.CreateInferredEventCallback}(this, __value => {original.Content} = __value, {original.Content})",
+            Content = $"{ComponentsApi.RuntimeHelpers.CreateInferredEventCallback}(this, ",
+            Kind = TokenKind.CSharp
+        });
+
+        switch ((setter, after))
+        {
+            case (null, null):
+                // no bind:set nor bind:after, use the same code generation as before
```
SteveSandersonMS, line 868: Suggestion:

```
// no bind:set nor bind:after, so we simply assign to the bound expression and we're done
```

... since "before" won't make sense after the PR is merged :)

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -740,7 +975,52 @@ static CSharpExpressionIntermediateNode ExtractEventNodeExpression(TagHelperDire
...
         {
-            Content = $"global::{ComponentsApi.EventCallback.FactoryAccessor}.{ComponentsApi.EventCallbackFactory.CreateBinderMethod}(this, __value => {original.Content} = __value, ",
+            Content = $"global::{ComponentsApi.EventCallback.FactoryAccessor}.{ComponentsApi.EventCallbackFactory.CreateBinderMethod}(this, ",
+            Kind = TokenKind.CSharp
+        });
+
+        switch ((setter, after))
+        {
+            case (null, null):
+                // no bind:set nor bind:after, use the same code generation as before
```
SteveSandersonMS, line 985: Suggestion:

```
// no bind:set nor bind:after, so we simply assign to the bound expression and we're done
```

... since "before" won't make sense after the PR is merged :)

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -740,7 +975,52 @@ static CSharpExpressionIntermediateNode ExtractEventNodeExpression(TagHelperDire
...
+                // bind:set and bind:after create the code even though we disallow this combination through a diagnostic
+                var setToEventCallback = $"{ComponentsApi.RuntimeHelpers.CreateInferredEventCallback}(this, callback: {setter.Content}, value: {original.Content})";
+                afterToEventCallback = $"global::{ComponentsApi.EventCallback.FactoryAccessor}.{ComponentsApi.EventCallbackFactory.CreateMethod}(this, callback: {after.Content})";
+                changeExpressionTokens.Add(new IntermediateToken()
+                {
+                    Content = $"{ComponentsApi.RuntimeHelpers.CreateInferredEventCallback}(this, callback: async __value => {{ await {setToEventCallback}.InvokeAsync(); await {afterToEventCallback}.InvokeAsync(); }}, value: {original.Content})",
+                    Kind = TokenKind.CSharp
+                });
+                break;
+        }
```
SteveSandersonMS, line 1019: BTW there's a lot of logic that seems to be replicated 3 times in this file. Is there any way of factoring it out, or are there subtle differences that make this impractical? Just thinking of the risk of subtle differences being added in the future.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -826,7 +1106,8 @@ private class BindEntry
         public BindEntry(IntermediateNodeReference bindNodeReference)
         {
             BindNodeReference = bindNodeReference;
-            BindNode = (TagHelperDirectiveAttributeIntermediateNode)bindNodeReference.Node;
+            BindNode = bindNodeReference.Node as TagHelperDirectiveAttributeIntermediateNode;
```
SteveSandersonMS, line 1109: Is there a reason why the cast is no longer valid? If it is still valid, would be good to stick with that to avoid giving the impression that we have to deal with the types not matching at all.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentMetadata.cs
```diff
@@ -91,6 +93,8 @@ public static class Component
 
         public const string DelegateSignatureKey = "Components.DelegateSignature";
 
+        public const string DelegateAwaitableResultKey = "Components.DelegateAwaitableResult";
```
SteveSandersonMS, line 96: I think this is used as a boolean flag. If that's the case, could we follow the same naming convention as `IsWeaklyTyped` and name it `IsDelegateWithAwaitableResult` or similar?

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentsApi.cs
```diff
@@ -109,6 +109,8 @@ public static class RuntimeHelpers
     {
         public const string TypeCheck = "global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.TypeCheck";
         public const string CreateInferredEventCallback = "global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.CreateInferredEventCallback";
+        public const string InvokeSynchronousDelegate = "global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.InvokeSynchronousDelegate";
+        public const string InvokeAsynchronousDelegate = "global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.InvokeAsynchronousDelegate";
```
SteveSandersonMS, line 113: I'd be interested to see how these helpers will look (e.g., their range of overloads). Does that info exist anywhere yet?

src/Microsoft.AspNetCore.Razor.Language/src/Components/TagHelperDescriptorExtensions.cs
```diff
@@ -44,7 +44,7 @@ public static bool IsInputElementBindTagHelper(this TagHelperDescriptor tagHelpe
     {
         return
             tagHelper.IsBindTagHelper() &&
-            tagHelper.TagMatchingRules.Count == 1 &&
+            tagHelper.TagMatchingRules.Count == 2 &&
```
SteveSandersonMS, line 47: I know this caused you some inconvenience. Do you think it would work for us to remove this check, or does it serve some important purpose?

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentRuntimeCodeGenerationTest/EventHandler_WithDelegate_PreventDefault/TestComponent.codegen.cs
```diff
@@ -30,7 +30,7 @@ protected override void BuildRenderTree(global::Microsoft.AspNetCore.Components.
 #line hidden
 #nullable disable
             ));
-            __builder.AddEventPreventDefaultAttribute(2, "onfocus", 
+            __builder.AddEventPreventDefaultAttribute(2, "onfocus:preventDefault",
```
SteveSandersonMS, line 33: Has this been broken?

src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs
```diff
@@ -1051,7 +1051,7 @@ public class ComponentWithEditorRequiredChildContent : ComponentBase
     #region Bind
 
     [Fact]
-    public void BindToComponent_SpecifiesValue_WithMatchingProperties()
+    public void BindToComponent_WithMatchingProperties()
```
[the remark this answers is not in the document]
SteveSandersonMS: Actually it would be ideal if you could revert any of these test case name changes, because the name changes make the baseline diffs impractical to review. If you think the name changes are valuable, would it be possible to re-do them when this PR is done and code review is complete?

src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentRuntimeCodeGenerationTest/BindToComponent_EventCallback_TypeChecked_WithMatchingProperties/TestComponent.ir.txt
```diff
@@ -13,6 +13,8 @@ Document -
                             LazyIntermediateToken - (26:0,26 [11] x:\dir\subdir\Test\TestComponent.cshtml) - CSharp - ParentValue
                     ComponentAttribute - (26:0,26 [11] x:\dir\subdir\Test\TestComponent.cshtml) - ValueChanged - ValueChanged - AttributeStructure.DoubleQuotes
                         CSharpExpression - 
-                            IntermediateToken -  - CSharp - global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.CreateInferredEventCallback(this, __value => ParentValue = __value, ParentValue)
+                            IntermediateToken -  - CSharp - global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.CreateInferredEventCallback(this, 
+                            IntermediateToken -  - CSharp - __value => ParentValue = __value
+                            IntermediateToken -  - CSharp - , ParentValue)
```
SteveSandersonMS, line 18: I'm not sure if breaking this over multiple lines is for an important reason, but it does add a lot of noise to the baseline diffs.

Is there any chance you could put this part of the code formatting back to how it was before, to minimize baseline churn, and then maybe do a separate PR (or at least an independent commit for it at the end of this PR after the code review is done)?

SteveSandersonMS:
So far, this looks like a great enhancement!

src/Microsoft.CodeAnalysis.Razor/src/BindTagHelperDescriptorProvider.cs
```diff
@@ -196,6 +197,35 @@ private TagHelperDescriptor CreateFallbackBindTagHelper()
...
+                parameter.Documentation = ComponentResources.BindTagHelper_Element_Get_Documentation;
+
+                parameter.SetPropertyName("Get");
+
+                parameter.SetBindAttributeAlternative();
```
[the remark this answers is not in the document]
javiercn: The compiler tries to be "agnostic" of these things (arguably flawed) but I'll be explicit about it.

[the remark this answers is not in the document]
javiercn: In the beginning I thought this would accept a function, but it needs to be a named value to compile (unless we wanted to change that to allow providing a function, which I don't think we want).

src/Microsoft.CodeAnalysis.Razor/src/BindTagHelperDescriptorProvider.cs
javiercn: In the beginning I thought this would accept a function, but it needs to be a named value to compile (unless we wanted to change that to allow providing a function, which I don't think we want).

src/Microsoft.CodeAnalysis.Razor/src/ComponentTagHelperDescriptorProvider.cs
```diff
@@ -261,6 +263,57 @@ bool HasTypeParameter(ITypeSymbol type)
         }
     }
 
+    private static string IsAwaitable(IPropertySymbol prop)
```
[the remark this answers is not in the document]
javiercn: We need for component bindings in the form of  Value and ValueChanged where ValueChanged can be `Action<T>` or `Func<T, Task>`. 

Unlike for other bindings when we can always create an EventCallback, in this case we can't (in the synchronous case at least). So what this method does is, when we are inspecting the ValueChanged property, (which we already know is a delegate) we check to see if it has a return type, and if that return type is `awaitable`, which covers Task, ValueTask, and any other potential awaitable in the future.

The logic in this method is the same that the roslyn compiler uses to make that determination.

src/Microsoft.CodeAnalysis.Razor/test/BindTagHelperDescriptorProviderTest.cs
javiercn: That's a good catch, I'll add some additional assertions.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
javiercn: Unfortunately the two nodes don't share a common interface, that's why all the null coalescing and ternary operators going around. I didn't want to change the shape of the intermediate nodes by adding a common interface, as I didn't want to "hide" the actual types that I was dealing with, hence all the "duplication".

That said, I'll check if I can find a reasonable way of factoring the code, maybe inside BindEntry so these things can be shared.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
javiercn: I agree, but I also didn't want to make such an impactful change as part of this change. I believe that we should tackle this separately and with a clear goal, rather than performing ad-hoc refactoring.

I'll add explanatory comments as well as a more detailed write-up of the change so that we can keep track of it.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
javiercn: Yeah, that was part of my follow-up plan on cleaning this area. I want to avoid major refactors on the tree/IR as part of this change, but something local like moving some of the logic to BindEntry will work.

Unfortunately something like EffectiveGetNode won't work because the common ancestor for the types is IntermediateNode, which doesn't have access to the required properties.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
javiercn: I don't feel strongly about it. I prefer a switch statement because it can be read as a truth table and makes it easier to make sure we are dealing with all cases, but as I said, I don't mind changing it, so will do so.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentsApi.cs
javiercn: https://github.com/dotnet/aspnetcore/pull/40143/files#diff-75ff7c1daf5ea067e9a29e6ddba5450cf62daab8ff131de1ed77bd2816478de7R55-R126

src/Microsoft.AspNetCore.Razor.Language/src/Intermediate/ComponentAttributeIntermediateNode.cs
```diff
@@ -97,13 +97,13 @@ public ComponentAttributeIntermediateNode(TagHelperDirectiveAttributeParameterIn
             throw new ArgumentNullException(nameof(directiveAttributeParameterNode));
         }
 
-        AttributeName = directiveAttributeParameterNode.AttributeNameWithoutParameter;
+        AttributeName = directiveAttributeParameterNode.AttributeName;
```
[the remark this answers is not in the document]
javiercn: This was unintended, I reverted it.

src/Microsoft.AspNetCore.Razor.Language/src/Components/TagHelperDescriptorExtensions.cs
javiercn: We could make it check that there are at least 1 rules, but its going to be very rare that this changes again, so I'm ok with living with this unfortunate bit and just keeping it as it.

src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs
```diff
@@ -1242,7 +1242,7 @@ public class MyComponent : ComponentBase
     }
 
     [Fact]
-    public void BindToComponent_SpecifiesValue_WithoutMatchingProperties()
+    public void BindToComponent_WithoutMatchingProperties()
```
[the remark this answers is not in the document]
javiercn: A bit off topic, but `@bind` is just syntactic sugar. It doesn't check that there is a matching property in the same way parameters aren't checked unless you add something like [EditorRequired] to them.

Keep in mind that your component might not extend ComponentBase at all, and this could be a valid binding still.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDiagnosticFactory.cs
```diff
@@ -536,4 +535,18 @@ public static RazorDiagnostic CreateBindAttributeParameter_InvalidSyntaxBindSetA
         return diagnostic;
     }
 
+    public static readonly RazorDiagnosticDescriptor BindAttributeParameter_UnsupportedSyntaxBindGetSet =
+        new RazorDiagnosticDescriptor(
+        $"{DiagnosticPrefix}10020",
+        () => "Attribute '{0}' can't only be used with RazorLanguageVersion 7.0 or higher.",
```
SteveSandersonMS, line 541:
```suggestion
        () => "Attribute '{0}' can only be used with RazorLanguageVersion 7.0 or higher.",
```

src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs
```diff
@@ -1613,16 +1621,23 @@ public void BindToElementWithSuffix_WritesAttributes()
...
+    [BindElement(""div"", ""myvalue"", ""myvalue"", ""myevent"")]
     public static class BindAttributes
     {
     }
 }"));
+
         // Act
         var generated = CompileToCSharp(@"
-<div @bind-value=""@ParentValue"" />
+<div @bind-myvalue:get=""@ParentValue"" @bind-myvalue:set=""ValueChanged"">
```
SteveSandersonMS, line 1632: The diff in this file is basically unreadable as it's matching up unrelated old tests with new ones. Not sure I'm going to be able to reason about this.

SteveSandersonMS (approved):
To be honest I think this is beyond what I can realistically review:

 * Partly because the nature of how the Razor compiler works involves a lot of action-at-a-distance so you can't really see how things work in one place for such a broad change
 * Partly because the diffs around tests are messed up and falsely match old tests with new ones

I did just skim read through a lot of the new implementation code, but didn't glean much insight from it. However, I can see you've done a lot of hard thinking on how difference sub-cases should be handled, and I have no reason to specifically think anything wrong. So, approving to make sure you're unblocked.

Ideally I'd love to see maybe about 5 examples of the different shapes of runtime code that can get generated now (for different combinations of get/set/after, and whether or not it's async). If you happen to be able to point to any representative examples in the baselines I'd be keen to have a look. But again, I know we've discussed the design at some length in the past and it made sense to both of us, so if you're saying there weren't any worrying surprises during implementation I'm happy to go with it!

mckaragoz:
Is there any article that explains a bit more? Don't we need to call ValueChanged anymore, or we will call in after method?

boukenka:
@mckaragoz Please have a look at [ASP.NET Core Blazor data binding](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?view=aspnetcore-7.0)

PR #73, opened 2022-02-15
[Blazor] Avoids parsing type names and caching results which might cause null reference exception

javiercn:
This PR avoids parsing type names for TagHelperDescriptors (mainly used by components) and instead relies on Roslyn during tag helper discovery to compute and save the component namespace as well as the type name identifier. The type name identifier reflects the type name without any generic arguments.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/src/DefaultRazorTagHelperBinderPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/src/DefaultRazorTagHelperBinderPhase.cs
@@ -265,8 +265,7 @@ internal class DefaultRazorTagHelperBinderPhase : RazorEnginePhaseBase, IRazorTa
                 {
                     // If this is a child content tag helper, we want to add it if it's original type is in scope.
                     // E.g, if the type name is `Test.MyComponent.ChildContent`, we want to add it if `Test.MyComponent` is in scope.
-                    TrySplitNamespaceAndType(tagHelper, out var typeNamespace);
-                    if (!typeNamespace.IsEmpty && IsTypeInScope(typeNamespace, currentNamespace))
+                    if (IsTypeInScope(tagHelper, currentNamespace))
                     {
                         Matches.Add(tagHelper);
                     }
@@ -352,8 +351,7 @@ internal class DefaultRazorTagHelperBinderPhase : RazorEnginePhaseBase, IRazorTa
                         {
                             // If this is a child content tag helper, we want to add it if it's original type is in scope of the given namespace.
                             // E.g, if the type name is `Test.MyComponent.ChildContent`, we want to add it if `Test.MyComponent` is in this namespace.
-                            TrySplitNamespaceAndType(tagHelper, out var typeName);
-                            if (!typeName.IsEmpty && IsTypeInNamespace(typeName, @namespace))
+                            if (IsTypeInNamespace(tagHelper, @namespace))
                             {
                                 Matches.Add(tagHelper);
                             }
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNetCore.Razor.Language/src/DefaultRazorTagHelperBinderPhase.cs
```diff
@@ -368,63 +366,38 @@ public override void VisitRazorDirective(RazorDirectiveSyntax node)
...
                 // Either the typeName is not the full type name or this type is at the top level.
-                return true;
+                return string.IsNullOrEmpty(typeNamespace) || typeNamespace.Equals(@namespace, StringComparison.Ordinal);
             }
-
-            return typeNamespace.Equals(@namespace, StringComparison.Ordinal);
         }
 
+        // Just for unit testing backwards compatibility with the previous implementation
+
```
javiercn, line 381:
```suggestion
```

src/Microsoft.AspNetCore.Razor.Language/src/DefaultRazorTagHelperBinderPhase.cs
```diff
@@ -445,83 +418,7 @@ private static bool IsTypeInScopeCore(string currentNamespace, StringSegment typ
...
-            @namespace = fullTypeName.Subsegment(0, splitLocation);
-
-            var typeNameStartLocation = splitLocation + 1;
-            if (typeNameStartLocation < fullTypeName.Length)
-            {
-                typeName = fullTypeName.Subsegment(typeNameStartLocation, fullTypeName.Length - typeNameStartLocation);
-            }
-
-            return true;
+            return ComponentMetadata.IsMangledClass(tagHelper.GetTypeNameIdentifier());
```
javiercn, line 421: >                // If this is a child content tag helper, we want to look at it's original type.
>                // E.g, if the type name is `Test.__generated__MyComponent.ChildContent`, we want to look at `Test.__generated__MyComponent`.

This comment from code no longer applies since the previous code was parsing the FullTypeName and for some reason, we added the property name to child component tag helpers. Since we no longer look at that, and use a different property for this checks, I don't append the property name to them, so we can check the namespace directly.

src/Microsoft.AspNetCore.Razor.Language/test/DefaultRazorTagHelperBinderPhaseTest.cs
```diff
@@ -1344,57 +1350,6 @@ public void IsTagHelperFromMangledClass_WorksAsExpected()
...
-
-        // Try again to make sure caching works
-        result = DefaultRazorTagHelperBinderPhase.ComponentDirectiveVisitor.TrySplitNamespaceAndType(
-            tagHelperDescriptor, out @namespace, out typeName);
-
-        Assert.Equal(expectedResult, result);
-        Assert.True(new StringSegment(expectedNamespace).Equals(@namespace, StringComparison.Ordinal));
-        Assert.True(new StringSegment(expectedTypeName).Equals(typeName, StringComparison.Ordinal));
-    }
-
```
javiercn, line 1397: These tests no longer apply since we don't do no parsing no more.

src/Microsoft.CodeAnalysis.Razor/src/EventHandlerTagHelperDescriptorProvider.cs
```diff
@@ -250,7 +258,8 @@ private struct EventHandlerData
         public string Assembly { get; }
 
         public string TypeName { get; }
-
+        public string TypeNamespace { get; }
+        public string TypeNameIdentifier { get; }
```
javiercn, line 262:
```suggestion
        public string TypeNamespace { get; }

        public string TypeNameIdentifier { get; }

```

src/Microsoft.CodeAnalysis.Razor/test/DefaultTagHelperDescriptorFactoryTest.cs
```diff
@@ -999,6 +1067,8 @@ public void CreateDescriptor_OverridesAttributeNameFromAttribute()
                 "overridden-attribute",
                 typeof(OverriddenAttributeTagHelper).FullName,
                 AssemblyName,
+                            typeNamespace: typeof(OverriddenAttributeTagHelper).FullName.Substring(0, typeof(OverriddenAttributeTagHelper).FullName.Length - nameof(OverriddenAttributeTagHelper).Length -1),
+                            typeNameIdentifier: nameof(OverriddenAttributeTagHelper),
```
javiercn, line 1071:
```suggestion
                typeNamespace: typeof(OverriddenAttributeTagHelper).FullName.Substring(0, typeof(OverriddenAttributeTagHelper).FullName.Length - nameof(OverriddenAttributeTagHelper).Length -1),
                typeNameIdentifier: nameof(OverriddenAttributeTagHelper),
```

PR #76, opened 2022-02-17
Normalize source files

sharwell:
Fix a non-normalized source file checked in.

```
git add --all --renormalize
```


TanayParikh (approved):
Thanks!

PR #77, opened 2022-02-17
Normalize source files (7.0 preview 2)

sharwell:
Fix a non-normalized source file checked in.

```
git add --all --renormalize
```


TanayParikh (approved):

dougbu (approved):

dougbu:
> Why is it necessary to take this into Preview 2?

Because w/o the change, official builds fail about half the time. No idea why builds don't fail _all_ the time.

PR #85, opened 2022-03-01
Update branding to 6.0.4

vseanreesermsft:


dougbu:
I may be wrong about needing branding updates here. @pranavkm what is the right policy for this branch❔

pranavkm:
So far we've been branching with the SDK. That said, I'm not super sure if we have any planned work for this 6.0.4xx, so it might not be necessary to continue doing this?

dougbu:
@pranavkm my point was about branding, not creating new branches. If we have been fine w/ non-stable branding and leaving the prerelease version alone, we shouldn't need PRs like this one. So, have we done anything similar since branching release/6.0.2xx❔

If not, we can close this PR and #83. My thinks to @vseanreesermsft -- especially if this was unnecessary.

pranavkm:
>  If we have been fine w/ non-stable branding and leaving the prerelease version alone, we shouldn't need PRs like this one.

I would think so. @TanayParikh / @javiercn in case you feel differently.

TanayParikh:
> > If we have been fine w/ non-stable branding and leaving the prerelease version alone, we shouldn't need PRs like this one.
> 
> I would think so. @TanayParikh / @javiercn in case you feel differently.

Yeah given we aren't planning any immediate work here, I think it may be fine to leave it alone.

dougbu:
Thanks again @vseanreesermsft

PR #88, opened 2022-03-08
[Blazor] Avoid redundantly qualifying event callback type checks

javiercn:
There were a couple of issues here:
* The  code that parsed EventCallback type arguments didn't account for `global::`.
* Once that got fixed, the EventCallback code wasn't distinguishing between explicit types and implicit/inferred types.
* For inferred types, there were two cases:
  * One where the type was replaced by a user defined type. In this case, we need to call `CreateEventCallback<Type>`
  * Two where the type was not replaced and needed to be inferred. In this case we need to call `CreateEventCallback` because the argument is being inferred using `CreateInferredEventCallback`

The change uses properties to track whether the user provided a type explicitly in code and whether we were able to replace the generic type properties in EventCallback and uses those facts to influence code generation.


captainsafia (approved):

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
```diff
@@ -738,18 +745,26 @@ private void WriteComponentAttributeInnards(CodeRenderingContext context, Compon
                 }
             }
 
-            static void QualifyEventCallback(CodeWriter codeWriter, string typeName)
+            static void QualifyEventCallback(CodeWriter codeWriter, string typeName, bool? explicitType)
             {
-                if(ComponentAttributeIntermediateNode.TryGetEventCallbackArgument(typeName, out var argument))
+                if (ComponentAttributeIntermediateNode.TryGetEventCallbackArgument(typeName, out var argument))
                 {
+                    codeWriter.Write("global::");
```
TanayParikh, line 752: Given the issue is with redundant globals (and that's covered in 755-762 directly below), why is this additional global for the event callback type name now required?

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentGenericTypePass.cs
```diff
@@ -358,10 +360,18 @@ private void RewriteTypeNames(TypeNameRewriter rewriter, ComponentIntermediateNo
...
                     // acount for that during code generation and avoid trying to fully qualify
                     // the type name.
-                    if (!globallyQualifiedTypeName.StartsWith("global::", StringComparison.Ordinal))
+                    if (hasTypeArgumentSpecified == true)
                     {
                         attribute.Annotations.Add(ComponentMetadata.Component.ExplicitTypeNameKey, true);
                     }
+                    else if(attribute.BoundAttribute?.IsEventCallbackProperty() ?? false)
+                    {
+                        var callbackArgument = typeNameFeature.ParseTypeParameters(attribute.TypeName)[0];
```
TanayParikh, line 369: Any way `typeNameFeature.ParseTypeParameters(attribute.TypeName)` can be null/empty?

TanayParikh (approved):
LGTM, just some clarifying questions for my understanding.

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentDesignTimeNodeWriter.cs
javiercn: Because we are writing the eventcallback type `by hand` see that `ComponentsApi.EventCallback.FullTypeName` writes `Microsoft.AspNetCore.Components.EventCallback`

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentGenericTypePass.cs
javiercn: No, because at this point we already know we have an EventCallback and the property is generic, so it means we have an EventCallback<T>

PR #214, opened 2022-04-15
[main] Resolve source-build prebuilts

lbussell:
Razor-compiler is being onboarded to source-build. Since we don't run repo tests in source-build, we need to exclude test projects so we don't download pre-built packages for the testing framework. See the following for more information:

- https://github.com/dotnet/source-build/issues/2825
- https://github.com/dotnet/installer/pull/13652 and https://github.com/dotnet/installer/pull/13631.

If these changes get merged I will submit a PR to backport them to `6.0.3xx` as well.

### DIFF
--- a/src/tools/RazorSyntaxGenerator/RazorSyntaxGenerator.csproj
+++ b/src/tools/RazorSyntaxGenerator/RazorSyntaxGenerator.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Description>Generates Razor syntax nodes from xml. For internal use only.</Description>
     <TargetFramework>netcoreapp3.1</TargetFramework>
-    <TargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net7.0</TargetFramework>
+    <TargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net6.0</TargetFramework>
     <AssemblyName>dotnet-razorsyntaxgenerator</AssemblyName>
     <PackageId>RazorSyntaxGenerator</PackageId>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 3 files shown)


MichaelSimons:
Friendly ping @https://github.com/orgs/dotnet/teams/aspnet-blazor-eng, can someone please review this to unblock source-build.

src/tools/RazorSyntaxGenerator/RazorSyntaxGenerator.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <Description>Generates Razor syntax nodes from xml. For internal use only.</Description>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net6.0</TargetFramework>
```
SteveSandersonMS, line 6: Could you clarify why this line is needed?

Apologies if this seems like being awkward, but I'm unsure why we'd be specifying `net6.0` here when `main` is now built on .NET 7. I can't guess whether it's right like this, or whether we'd ever change it in the future, without knowing what necessitates having this line at all. Hope that's OK!

lbussell: Yes, my bad, I initially made these changes to for 6.0.3xx source build so I forgot to change this target framework when I submitted this PR. The change prevents us from downloading .NET Core 3.1 packages (prebuilts) during source build.

wtgodbe (approved):

src/tools/RazorSyntaxGenerator/RazorSyntaxGenerator.csproj
lbussell: On closer inspection, this change is no longer needed since it does not pull in prebuilts in the context of .NET 7.0 source build.

lbussell:
@wtgodbe @SteveSandersonMS I don't have permissions to merge this, so if the change is acceptable, please merge when you can

PR #251, opened 2022-05-31
Remove ConditionalWeakTable usage

Erarndt:
The lifetime of the diagnostics and annotations collections were being bound to the lifetime of a specific GreenNode instance by adding it to a ConditionalWeakTable. However, this ends up being no different than making these items instance fields, and the large number of dependent handles internally created by the ConditionalWeakTable has a measurable negative impact on garbage collection performance.

For comparison, I took some traces using OrchardCore before and after the changes to see what the impact would be. Since this was on my local machine, there is room for error, but there is a very clear difference between the two traces

Before:

image

After:

image

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/src/Syntax/GreenNode.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/src/Syntax/GreenNode.cs
@@ -9,7 +9,6 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Runtime.CompilerServices;
 using System.Text;
 
 namespace Microsoft.AspNetCore.Razor.Language.Syntax;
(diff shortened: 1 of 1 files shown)


333fred:
This is basically the same change that @sharwell proposed in https://github.com/dotnet/roslyn/pull/57643, which we did not accept due to the increased size of GreenNode.

As another note, in general I do not want to modify the implementation of `GreenNode` in this repository right now. We are going to be working towards general razor compiler improvements, and one of my goals with that will be to share `GreenNode` implementations between the razor compiler and roslyn. After that is completed, we can consider optimizations around CWT that would affect all users of this code, rather than causing divergence.

333fred:
That is one of the reasons for it.

As for the timeframe, short to medium term. We'll be looking at both short-term improvements and general rearchitectures to address engineering debt.

PR #267, opened 2022-06-16
Ensure the Razor generator doesn't generate when it's suppressed

jasonmalinowski:
https://github.com/dotnet/sdk/pull/24928 made a change to the Razor generator to ensure that even if the generator is suppressed, it's still going to run and cache it's outputs so later runs aren't a from-scratch run on a performance-critical path. However, a bug meant that if the generator was suppressed, it'd still run the first time it's invoked, and will still output files even though it was supposed to be suppressed.

The approach taken here is to suppress the addition of the files, but at the very end of the chain only -- the generation and walking in the middle of the incremental chain is left untouched, so that still has the existing caching behavior.

This fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1545938 properly; https://github.com/dotnet/roslyn/pull/61675 was put in as a tactical fix first.

### DIFF
--- a/src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
+++ b/src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
@@ -233,9 +233,17 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
                     return string.Equals(a.csharpDocument.GeneratedCode, b.csharpDocument.GeneratedCode, StringComparison.Ordinal);
                 }, static a => StringComparer.Ordinal.GetHashCode(a.csharpDocument));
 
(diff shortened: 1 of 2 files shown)


chsienki (approved):

chsienki:
The changes to the SDK version got merged over here. We don't have the specific end result change, but I believe its the same effect. Happy to add in the 'just in case' bail out too though, although the source got moved around again (sorry) so will need a rebase.

jasonmalinowski:
@chsienki Yeah, no reason to have a duplicate check I don't think. I had missed #292 when looking for the PR that updated this repo, so I'm going to close this then.

PR #273, opened 2022-06-16
Remove unnecessary hashcode components for taghelpers

333fred:
Analysis of the traces from https://developercommunity.visualstudio.com/t/Slow-build-times-with-Blazor/10059673 show a large amount of time spent in hashcode calculation. Tag helpers were using hashcode as a true equality check, which caused a large amount of time to be spent in hashcode calculation that isn't necessary. https://github.com/dotnet/razor-tooling/pull/6515 pulls that code into where it should be, cache invalidation, and the compiler can then remove the unnecessary hashcode work.


333fred:
@chsienki @RikkiGibson for review.

jaredpar:
Does razor-tooling have to think about what happens when older generators / compilers are loaded into the VS process? Basically is there a situation where a 17.2 compiler can be loaded into 17.3 VS? I know the reverse is true today.

333fred:
I really hope the answer is no, but I don't know for certain.

333fred:
> Yes :( If is select an older SDK then the generator (and thus compiler) that is used will be from that older SDK. As an aside this is something we're going to have to be really careful about if we want to start shipping the generator in tooling as well as the SDK.

So does this mean we actually can't take this change, because it would break older versions of the tooling?

jaredpar:
@chsienki 

> Yes :( If is select an older SDK then the generator (and thus compiler) that is used will be from that older SDK

How can you select an older SDK though? Generally VS allows .NET SDK to move forward but not backward. Not sure the scenario by which an older one can be chosen.

chsienki:
Because the generator is loaded from the SDK that is used to build it. So `global.json` means I can change which generator is loaded into VS. 

Now, thinking about it the tooling doesn't actually *use* that generator right now, it's compiled with its own copy of the compiler so the design time stuff should actually be fine. I think that hot reload should also be ok as its just switching to the generator.

So maybe we're ok here?

jaredpar:
I'm pretty confident that loading a SDK older than the version of VS is strictlry unsupported. Let's verify though.

NTaylorMullen:
Tooling does need to care about older Razor compiler / SG versions; however, it still utilizes the same TagHelper types for all versions. Meaning, I don't think this will impact the multi-targeting capabilities specifically.

Now that being said, "will this break tooling": Possibly. If it does the only reason it will is that we missed a consumption point when Fred [moved the hash code calc logic into the tooling side](https://github.com/dotnet/razor-tooling/pull/6515). So I think we should go ahead with this change and react on the tooling side where necessary / keep an eye on if things break.

NTaylorMullen:
> What we're not certain about is older tooling that does not have my tooling change that observes a newer compiler that does have this hashcode change.

Should be ok. Older tooling doesn't actually load anything into VS. Razor tooling configures its Razor compiler in-proc based on what the MSBuild properties say the "razor version" is. Meaning, no matter what ASP.NET Core project version the Razor compiler dll will always be 7.X.X.X where 7 is "latest" and then the tooling will set various flags to tell the compiler to behave as if it was an older version.

333fred:
> and then the tooling will set various flags to tell the compiler to behave as if it was an older version.

Does that mean that I need to look for these flags here and go back to hold hashcode calculation if it's in that older version mode?

NTaylorMullen:
> Does that mean that I need to look for these flags here and go back to hold hashcode calculation if it's in that older version mode?

 No

jcouv (approved):
LGTM Thanks (iteration 3)

jcouv:
Just curious, did we find a way to deploy/reference local bits in a project? We could see how much this improved the perf and see what sticks  out next.

333fred:
> Just curious, did we find a way to deploy/reference local bits in a project? We could see how much this improved the perf and see what sticks out next.

I haven't been able to get the repro solution to build with the .NET 7.0 SDK, so I haven't been able to test this on it.=

333fred:
@dotnet/razor-compiler for a second review please.

RikkiGibson (approved):

Issue #213, opened 2022-04-15
SDK 7.0100-preview.3.22179.4 fails to create working razor.g.cs code for certain constructs. 


MarkStega:
**Describe the bug**

If you have this construct in a razor file in an RCL:
```
                        <MBChipsSelectMulti @bind-Value="@SBValues"
                                            Items="@KittenBreeds" />
```
then this line (155)  in the razor.g.cs file fails to compile:
```
                    , 20, global::Microsoft.AspNetCore.Components.EventCallback.Factory.Create<global::System.Collections.Generic.IList<TItem>>(this, global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.CreateInferredEventCallback(this, __value => SBValues = __value, SBValues)), 21, () => SBValues);
                    __builder3.CloseElement();
                    __builder3.AddMarkupContent(22, "\r\n\r\n                    ");
                    __builder3.OpenElement(23, "p");
                    global::__Blazor.Material.Blazor.Website.Pages.ChipsSelectMulti.TypeInference.CreateMBChipsSelectMulti_1(__builder3, 24, 25, 
```
with a errors of
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0246	The type or namespace name 'TItem' could not be found (are you missing a using directive or an assembly reference?)	Material.Blazor.Website	C:\Solutions\OHI\Material.Blazor\Material.Blazor.Website\Microsoft.NET.Sdk.Razor.SourceGenerators\Microsoft.NET.Sdk.Razor.SourceGenerators.RazorSourceGenerator\Pages_ChipsSelectMulti_razor.g.cs	155	N/A

Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1503	Argument 2: cannot convert from 'Microsoft.AspNetCore.Components.EventCallback<System.Collections.Generic.IList<string>>' to 'Microsoft.AspNetCore.Components.EventCallback'	Material.Blazor.Website	C:\Solutions\OHI\Material.Blazor\Material.Blazor.Website\Microsoft.NET.Sdk.Razor.SourceGenerators\Microsoft.NET.Sdk.Razor.SourceGenerators.RazorSourceGenerator\Pages_ChipsSelectMulti_razor.g.cs	155	N/A
```

**To Reproduce:**

Material.Blazor is an open source project available at https://github.com/Material-Blazor/Material.Blazor

You don't need to know anything about the project to reproduce the issue, simple:

[...]

Issue #100, opened 2022-04-07
[ 7.0.100-preview.3.22179.4] error CS1660 when trying to build a project that is using Telerik.UI.for.Blazor(2.22.0)


Junjun-zhao:
**Application Name**: Telerik.UI.for.Blazor(2.22.0) which is used by one another app Provided by our customer
**OS**: Windows 10 RS5
**CPU**: X64
**.NET Build Number**: 7.0.100-preview.3.22179.4

**App info checking at** : https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1520481

**Verify Scenarios:**

1). Windows rs5 x64 + 7.0.100-preview.3.22179.4: Fail
2). Windows rs5 x64 + dotnet-sdk-7.0.100-preview.1.22101.1: Pass
3). Windows rs5 x64 + dotnet-sdk-6.0.201: Pass

**Repro steps:**

We can copy app source to local machine, and build it with .net6 and .net7 SDKs, net7 will fail.(Please refer to the Source code 
[BlazorApp5.zip](https://github.com/dotnet/sdk/files/8442411/BlazorApp5.zip)

Minimal Repro steps:

1) Create blazor app (server side), any runtime is OK, our sample project is 5.0, but we will build it with 7.0.100-preview.3.22179.4
2) Add Telerik.UI.for.Blazor, our app is using 2.22.0 => <PackageReference Include="Telerik.UI.for.Blazor" Version="2.22.0" />
3) Add TelerikGrid to a page any page :  
```
   <TelerikGrid Data=@_projects  SelectedItemsChanged="@((IEnumerable<MyClass> ItemList) => OnSelect(_projects))">
                <GridColumns>
                      <GridColumn Field="@nameof(MyClass.MyProperty)" Title="Product Name" Width="195px" />
                 </GridColumns>
         </TelerikGrid>
```
4) Build it with 7.0.100-preview.3.22179.4,

**Expected Result:**

Build Successfully

**Actual Result:**
```
C:\Users\v-altunc\source\repos\BlazorApp5\BlazorApp5\obj\Debug\net5.0\Razor\Pages\Counter.razor.g.cs(112,134): error CS
0246: The type or namespace name 'TItem' could not be found (are you missing a using directive or an assembly reference
?) [C:\Users\v-altunc\source\repos\BlazorApp5\BlazorApp5\BlazorApp5.csproj]
[...]

PR #275, opened 2022-06-21
[Blazor] Fix issue with type inference and nested generic types

javiercn:
Fixes https://github.com/dotnet/razor-compiler/issues/213 and https://github.com/dotnet/razor-compiler/issues/100

The issue was that we were only looking at the first generic type parameter for the EventCallback to determine if there were generic type parameters to take into account.

As a result, on IEnumerable<T> we were not discovering the generic type parameter and we were marking issuing the wrong code.

This change iterates over the results of typeNameFeature.ParseTypeParameters which already flattens all the types contained in the signature.

Added a couple of tests to cover using an EventCallback directly and using an event callback via `@bind`

Also increased the coverage for multiple levels of inference (Generic type being consumed by another generic type and using an event callback with and without type inference)

The second commit also addresses https://github.com/dotnet/razor-compiler/issues/219 which was caused by a "naive" globalization approach instead of relying on the globallyqualifiedtype available.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentDesignTimeCodeGenerationTest/ChildComponent_NonGenericParameterizedChildContent_TypeInference/TestComponent.codegen.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/test/TestFiles/IntegrationTests/ComponentDesignTimeCodeGenerationTest/ChildComponent_NonGenericParameterizedChildContent_TypeInference/TestComponent.codegen.cs
@@ -63,7 +63,7 @@ namespace __Blazor.Test.TestComponent
     #line hidden
     internal static class TypeInference
     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentGenericTypePass.cs
```diff
@@ -366,10 +366,15 @@ private void RewriteTypeNames(TypeNameRewriter rewriter, ComponentIntermediateNo
                     }
                     else if(attribute.BoundAttribute?.IsEventCallbackProperty() ?? false)
                     {
-                        var callbackArgument = typeNameFeature.ParseTypeParameters(attribute.TypeName)[0];
-                        if (bindings.ContainsKey(callbackArgument))
+                        var typeParameters = typeNameFeature.ParseTypeParameters(attribute.TypeName);
```
javiercn, line 369: ParseTypeParameters already recurses over the TypeName and flattens the types parameters in the hierarchy, so for example for `Optional<IEnumerable<T>>` returns `[IEnumerable<T>, T]`.

TanayParikh (approved):
LGTM

PR #276, opened 2022-06-22
Temporarily produce Microsoft.AspNetCore.Razor.SourceGenerator.Tooling.Internal package

javiercn:
I believe this is necessary as the SDK is not consuming the new `Microsoft.NET.Sdk.Razor` transport package


javiercn:
Thanks @chsienki!

PR #244, opened 2022-05-20
Merge generator

chsienki:
Moves the razor source generator into this repo.

Some of the integration tests are still failing, so have been removed from the SLN while we bring them back up.

PR #287, opened 2022-06-29
Undo some merge

chsienki:
Remove a lot of the SDK stuff that was bought in with #244 

It turns out that the dependencies in the SDK are more complicated than we thought, and we'll need to think through the layering more carefully. We want to keep the generator code, however, so rather than reverting the merge (which bought history in from other repos) this just deletes the files we don't need. If we ever want to bring them back we can restore them from our own history (Commit 1)

I moved the source generator to its own folder, as it no longer makes sense to have it in a RazorSdk folder (Commit 2)

I created a new transport package that contains the razor source generator and its dependencies so we can ship it to both the SDK and the tooling repo (Commit 3)


chsienki:
@RikkiGibson and @333fred for an even bigger grim reaper PR.

src/Microsoft.NET.Sdk.Razor.SourceGenerators.Transport/Microsoft.NET.Sdk.Razor.SourceGenerators.Transport.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <IsShipping>false</IsShipping>
+    <SuppressDependenciesWhenPacking>false</SuppressDependenciesWhenPacking>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <GenerateDependencyFile>false</GenerateDependencyFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.NET.Sdk.Razor.SourceGenerators\Microsoft.NET.Sdk.Razor.SourceGenerators.csproj" />
+  </ItemGroup>
+
```
333fred, line 15: Extra whitespace

333fred (approved):

RikkiGibson (approved):

PR #288, opened 2022-07-01
Use Roslyn Assert diffing utilities in tests

333fred:
Today, Razor does baseline checking line-by-line, failing at the very first error. This makes it difficult to see the entire impact of a change to the complete baseline. Luckily, Roslyn already has a text-diffing assert algorithm, so I've copied that into the test base here and updated the baseline verifiers to use that, instead of splitting and manually validating a number of properties on each individual line.


src/test/Microsoft.AspNetCore.Razor.Test.Common/Assert/AssertEx.cs
```diff
@@ -0,0 +1,956 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
333fred, line 1: These files are copied nearly verbatim from roslyn. The only changes I made were to remove change the index behavior (as this project doesn't have `System.Range`), remove the `NotNull` assert (because this project doesn't have the nullable analysis attributes), and disable warnings.

333fred:
@RikkiGibson @chsienki for review.

src/test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxNodeVerifier.cs
```diff
@@ -3,295 +3,16 @@
...
-using Xunit;
-using Xunit.Sdk;
+using Roslyn.Test.Utilities;
 
 namespace Microsoft.AspNetCore.Razor.Language.Legacy;
 
 public static class SyntaxNodeVerifier
 {
-    internal static void Verify(SyntaxNode node, string[] baseline)
+    internal static void Verify(SyntaxNode node, string baseline)
```
jcouv, line 13: Seems to only be used in one place. Consider inlining or moving closer to usage.

src/test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/ClassifiedSpan/ClassifiedSpanVerifier.cs
```diff
@@ -3,110 +3,15 @@
...
-using Xunit;
-using Xunit.Sdk;
+using Roslyn.Test.Utilities;
 
 namespace Microsoft.AspNetCore.Razor.Language.Legacy;
 
 internal class ClassifiedSpanVerifier
 {
-    internal static void Verify(RazorSyntaxTree syntaxTree, string[] baseline)
+    internal static void Verify(RazorSyntaxTree syntaxTree, string baseline)
```
jcouv, line 12: Same comment (inline or move?)

jcouv (approved):
LGTM Thanks (iteration 1) with suggestion to remove the 3 tiny "verifier" files

jaredpar:
> @RikkiGibson @chsienki for review.

Note you can use @dotnet/razor-compiler for review purposes. It has all the right people attached to it

PR #289, opened 2022-07-01
Support patterns with nested colons in case statements

333fred:
The case statement parser looked until it found the first `:`. This isn't quite right: patterns can have nested `:` tokens. Instead, it needs to parse out recursive, tuple, and list patterns until it finds a `:` at the top level, and then it can consider the case clause complete. Fixes https://github.com/dotnet/razor-compiler/issues/220.


333fred:
@RikkiGibson @chsienki for review.

333fred:
@RikkiGibson @chsienki for reviews please.

src/test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/InitializeTestFileAttribute.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.AspNetCore.Razor.Language.IntegrationTests;
 
-public class IntializeTestFileAttribute : BeforeAfterTestAttribute
+public class InitializeTestFile : BeforeAfterTestAttribute
```
chsienki, line 11: Nit: keep the attribute postfix for the actual type name?

chsienki (approved):

333fred:
@RikkiGibson @dotnet/razor-compiler for a second review.

RikkiGibson (approved):

PR #308, opened 2022-07-26
Update SDK

333fred:
Move to using .NET 7p5, and language version preview.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,7 +11,7 @@
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <AnalysisLevel>latest</AnalysisLevel>
     <AnalysisMode>Default</AnalysisMode>
-    <LangVersion>Latest</LangVersion>
+    <LangVersion>preview</LangVersion>
     <Nullable>enable</Nullable>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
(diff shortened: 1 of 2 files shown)


Directory.Build.props
```diff
@@ -56,6 +56,10 @@
     <NoWarn>$(NoWarn.Replace('1591', ''))</NoWarn>
     <NoWarn Condition=" '$(BuildingInsideVisualStudio)' == 'true' ">$(NoWarn);0105</NoWarn>
 
+    <!-- Central package management requires that all packages have package source mapping, but we're using all dotnet feeds, and source-build and other downstream rebuilders of our souces
+         would need to support updating that list for their local builds. -->
+    <NoWarn>$(NoWarn);NU1507</NoWarn>
```
333fred, line 61: See https://docs.microsoft.com/en-us/nuget/consume-packages/package-source-mapping for info about mapping. I don't think it matters for our use case: we don't depend on any public feeds, only dnceng feeds, so picking up a maliciously-versioned preview from a public feed like nuget.org isn't an issue.

333fred:
@dotnet/razor-compiler for reviews.

chsienki (approved):

Directory.Build.props
```diff
@@ -11,7 +11,7 @@
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <AnalysisLevel>latest</AnalysisLevel>
     <AnalysisMode>Default</AnalysisMode>
-    <LangVersion>Latest</LangVersion>
+    <LangVersion>preview</LangVersion>
```
jcouv, line 14: I don't think we should use a preview LangVer in a shipped product, risking depending on a perview feature. 
Can we set this to 11 and upgrade it as needed?

jcouv: FWIW, it looks like we're using "preview" in roslyn (`settings.props`). Seems strange to me

jcouv (commented):
Done with review pass (iteration 2)

jcouv (approved):
LGTM Thanks (iteration 2)

Directory.Build.props
jcouv: Resolved offline

Issue #318, opened 2022-08-10
Blazor WebAssembly template does not compile in .NET 7.0 Preview 7


lbussell:
I only tested this on Linux (a Fedora 33 docker container). I don't know the behavior on Windows. Steps to reproduce:

1. Download .NET 7.0 preview 7 linux binaries from here: https://dotnet.microsoft.com/en-us/download/dotnet/thank-you/sdk-7.0.100-preview.7-linux-x64-binaries
2. Run `<path-to-dotnet-7.0p7>/dotnet new blazorwasm --name blazorwasm --output blazorwasm --language "C#"`
3. Run `<path-to-dotnet-7.0p7>/dotnet build`
4. Observe the error: `CS0246: The type or namespace name 'blazorwasm' could not be found (are you missing a using directive or an assembly reference?) [/data/blazorwasm/blazorwasm.csproj]`

Same issue as https://github.com/dotnet/source-build/issues/2946

lbussell:
I believe that this error happens because when I do a fresh clone of razor-compiler and run `./build.sh`, the resulting package `Microsoft.NET.Sdk.Razor.SourceGenerators.Transport.*.nupkg` is empty except for Icon.png. This is the package that is unpacked by the SDK here: https://github.com/dotnet/sdk/blob/360b1fc266cb098f0b526af07cab51c898e8ffe7/src/Layout/redist/targets/GenerateLayout.targets#L312

The entire `sdk/7.0.100-rc.1.22369.1/Sdks/Microsoft.NET.Sdk.Razor/source-generators/` folder is missing from the 7.0p7 SDK.

cc @MichaelSimons @crummel @333fred @chsienki @jaredpar

chsienki:
Investigating.

PR #319, opened 2022-08-11
Fix casing for transport content

chsienki:
Fixes https://github.com/dotnet/razor-compiler/issues/318 
Fixes https://github.com/dotnet/source-build/issues/2946


chsienki:
@dotnet/razor-compiler for review please

333fred (approved):

RikkiGibson (approved):

MichaelSimons:
@chsienki - How does this flow into the SDK repo?  I see the SDK depends on a [version](https://github.com/dotnet/sdk/blob/main/eng/Version.Details.xml#L205) from July.  I also noticed the SDK's razor-compiler dependencies are on multiple versions/commits (https://github.com/dotnet/sdk/blob/main/eng/Version.Details.xml#L218 and https://github.com/dotnet/sdk/blob/main/eng/Version.Details.xml#L214) which isn't good.

chsienki:
@MichaelSimons turns out the CI was failing due to component governance, so there is no new build to update to. I'm fixing that right now. 

Unsure how the dependencies got out of sync, but I'll put up an SDK pr as soon as the newer version is available. (I've checked with a darc update-dependencies that it brings them back into sync)

PR #340, opened 2022-09-01
Mitigation for tuple component parameters

333fred:
Mitigation for https://github.com/dotnet/razor-compiler/issues/332. We were generating code that looks like `global::(Tuple, Elements)`, which is illegal. In order to properly do this rewrite, we really need to parse the type name in `TypeNameHelper` and do a rewrite using a walk, but this is currently not possible because that layer does not have roslyn access. As a temporary mitigation until that is changed, we skip rewriting tuple types.


333fred:
@chsienki @dotnet/razor-compiler for review.

chsienki (approved):
LGTM. Can we also port it over to razor-tooling?

333fred:
> LGTM. Can we also port it over to razor-tooling?

I figured I would after this was merged.

333fred:
@dotnet/razor-compiler for a second review.

jaredpar (approved):

PR #363, opened 2022-10-07
[Blazor] Bind after changes

javiercn:
We changed how we want to approach this as part of fixing https://github.com/dotnet/aspnetcore/issues/43195 and the compiler needs to generate updated code.


src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
```diff
@@ -855,21 +855,21 @@ static CSharpExpressionIntermediateNode ExtractEventNodeExpression(TagHelperDire
...
         else
         {
             // bind:set and bind:after create the code even though we disallow this combination through a diagnostic
-            var setToEventCallback = $"{ComponentsApi.RuntimeHelpers.CreateInferredEventCallback}(this, callback: {setter.Content}, value: {original.Content})";
-            var afterToEventCallback = $"global::{ComponentsApi.EventCallback.FactoryAccessor}.{ComponentsApi.EventCallbackFactory.CreateMethod}(this, callback: {after.Content})";
+            var setToEventCallback = $"{ComponentsApi.RuntimeHelpers.CreateInferredBindSetter}(callback: {setter.Content}, value: {original.Content})";
+            var afterContentInvocation = $"{ComponentsApi.RuntimeHelpers.InvokeAsynchronousDelegate}(callback: {after.Content})";
             changeExpressionTokens.Add(new IntermediateToken()
             {
-                Content = $"{ComponentsApi.RuntimeHelpers.CreateInferredEventCallback}(this, callback: async __value => {{ await {setToEventCallback}.InvokeAsync(); await {afterToEventCallback}.InvokeAsync(); }}, value: {original.Content})",
```
jaredpar, line 872: What is the impact of removing `InvokeAsync`?

src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs
```diff
@@ -1933,7 +1933,7 @@ public class MyComponent : ComponentBase
     }
 
     [Fact]
-    public void BindToComponent_WithGetSet_EventCallback()
```
jaredpar, line 1936: Why was this test deleted? Deleting makes it harder to see the impact of this change because we can't see the comparison of old / new. It's a new test with new output.

mkArtakMSFT: Per the description of the issue, support for get/set/await was built on top of the existing EventCallback mechanism. However, this change changes that and doesn't use that mechanism for this new feature (bind get/set/await). Hence this test is not relevant anymore.

> This was due to the fact that we chose to rely on EventCallback to implement this feature and we did not realize there were some side effects associated with it.

> The APIs we are adding bypass the whole EventCallback infrastructure (simplifying the implementation too) and fix the issue.

mkArtakMSFT:
Tactics approved this change for .NET 7. @javiercn please work with the compiler team to get all the feedback addressed. Thanks!

src/Microsoft.AspNetCore.Razor.Language/src/Components/ComponentBindLoweringPass.cs
javiercn: The change is as follows:
* Before we would call `global::Microsoft.AspNetCore.Components.EventCallback.Factory.Create(this, callback: DoSomething)` which would create an `EventCallback` instance, and then call `InvokeAsync` on it.
* Now, we replaced that to use a different helper `InvokeAsynchronousDelegate` that receives the function/action and invokes it, so there is no need for InvokeAsync.
  * We already do this for situations where we are binding to a component that exposes a pair of Property and PropertyChanged properties, for example `string Value { get; set; }` and `Func<string, Task> ValueChanged { get; set; }` so this change makes make it similar to that case.

Here is the concrete code diff:
```diff
-            __o = global::Microsoft.AspNetCore.Components.EventCallback.Factory.CreateBinder(this, global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.CreateInferredEventCallback(this, callback: __value => { ParentValue = __value; return global::Microsoft.AspNetCore.Components.EventCallback.Factory.Create(this, callback: DoSomething).InvokeAsync(); }, value: ParentValue), ParentValue);
+            __o = global::Microsoft.AspNetCore.Components.EventCallback.Factory.CreateBinder(this, global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.CreateInferredBindSetter(callback: __value => { ParentValue = __value; return global::Microsoft.AspNetCore.Components.CompilerServices.RuntimeHelpers.InvokeAsynchronousDelegate(callback: DoSomething); }, value: ParentValue), ParentValue);
```

src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs
javiercn: This test was covering the case where an EventCallback instance was passed directly as the setter/after (which we introduced in the original implementation and causes problems), but there are tests for the other two cases already (function and action).

```csharp
BindToComponent_WithGetSet_Action()
BindToComponent_WithGetSet_ActionLambda()
BindToComponent_WithAfter_EventCallback_ReceivesAction()
BindToComponent_WithAfter_EventCallback_ReceivesFunction()
```

jaredpar: > Hence this test is not relevant anymore.

Deleting this test introduces substantial churn into the PR. It makes it substantially harder to read and harder to understand what did and didn't change. This is effectively a servicing level change, clarity needs to be prioritized. We should add these tests back so we can get a clearer understanding of the impact of this change.

javiercn: I brought back all the old tests and updated them to reflect that they will now fail compilation (as we do not support passing an EventCallback directly any longer) in addition to the existing additional tests that tests.

javiercn: In case it was not clear there is a very detailed explanation on the code generation code changes on the description of the PR, it was "hidden" behind a summary element, but I removed that just to make sure that its visible.

333fred (commented):
Done review pass (commit 6). One comment on tests, everything else is looking ok.

src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs
```diff
@@ -1957,14 +1955,17 @@ public class MyComponent : ComponentBase
...
 
         // Assert
         AssertDocumentNodeMatchesBaseline(generated.CodeDocument);
         AssertCSharpDocumentMatchesBaseline(generated.CodeDocument);
-        CompileToAssembly(generated);
+        var result = CompileToAssembly(generated, throwOnFailure: false);
+
+        var error = Assert.Single(result.Diagnostics);
+        // Argument 1: cannot convert from 'Microsoft.AspNetCore.Components.EventCallback' to '...' (System.Action/System.Func<Task>)
+        Assert.Equal("CS1503", error.Id);
```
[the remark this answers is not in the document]
javiercn: There is no generated code when it results in an error. Other tests in the area follow the same approach.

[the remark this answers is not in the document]
333fred: As far as I can tell, these are the only tests in this file that verify by ID. The rest have a diagnostics file in the baseline that verifies much more information about the diagnostic.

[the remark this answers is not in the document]
javiercn: What additional information would you want to verify?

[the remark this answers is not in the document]
javiercn: For reference, some other tests that follow a similar pattern
https://github.com/dotnet/razor-compiler/blob/fdbabdd098ec38adcba108d90895c86b68b460d5/src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs#L723-L727
https://github.com/dotnet/razor-compiler/blob/fdbabdd098ec38adcba108d90895c86b68b460d5/src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs#L1275-L1280
https://github.com/dotnet/razor-compiler/blob/fdbabdd098ec38adcba108d90895c86b68b460d5/src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs#L1347-L1352
https://github.com/dotnet/razor-compiler/blob/fdbabdd098ec38adcba108d90895c86b68b460d5/src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs#L1547-L1552
https://github.com/dotnet/razor-compiler/blob/fdbabdd098ec38adcba108d90895c86b68b460d5/src/Microsoft.AspNetCore.Razor.Language/test/IntegrationTests/ComponentCodeGenerationTestBase.cs#L2601-L2605

333fred (approved):
LGTM (commit 6), but we need to do a pass where we remove all of these verify only ID tests. These tests do not verify the end-user experience.

mkArtakMSFT (approved):

Issue #352, opened 2022-09-13
Support route syntax highlighting in Razor page directive


JamesNK:
The aspnetcore team is adding route syntax highlighting in .NET 8. It's currently working for routes specified in MVC attributes, e.g. `[HttpGet("page/{id}")]`, and minimal APIs, e.g. `app.MapGet("page/{id}", () => "")`. The final place where routes are defined is Razor.

The goal is to extend highlighting to work in a route specified with the Razor page directive. In the example screenshot below the route string parsed by the `RoutePatternFactory` has syntax highlighting applied. We want the route specified with the `@page` directive to get the same highlighting:

 ![image](https://user-images.githubusercontent.com/303201/189787360-61e4f353-45d6-4bb1-ad30-8c0770e30029.png)

The existing highlighting works via Roslyn detecting the StringSyntaxAttribute ([example with HttpGetAttribute](https://github.com/dotnet/aspnetcore/blob/c4bbe4b62b816dc36f7d1d5798c3a61e2bba5a17/src/Mvc/Mvc.Core/src/HttpGetAttribute.cs#L28)) and then running classifier code from the aspnetcore repo.

I believe the Razor compiler outputs the route in the `__typeHelper` field:

![image](https://user-images.githubusercontent.com/303201/189787527-a667a6b0-5574-4d62-917b-daf92518c90c.png)

If the `StringSyntax` attribute is added to the `__typeHelper` then Roslyn may just pick it up and automatically run the classifier for the string. Some testing will be required to verify this works, and potentially some further changes might be needed in Razor or Roslyn.

Also, standard ASP.NET Core routes used with Razor pages are slightly different from Blazor routes. The `StringSyntax` used for each should be different:

* Razor page: `StringSyntax("Route")`
* Blazor: `StringSyntax("BlazorRoute")`

cc @CyrusNajmabadi

JamesNK:
`__typeHelper` is a variable, not a field. An attribute won't work.

Instead, you could put a comment on the string to enable the classifier. For example:

![image](https://user-images.githubusercontent.com/303201/189890287-33212929-e7ab-41ef-b61b-05d71c71efb9.png)

davidwengier:
> `__typeHelper` is a variable, not a field. An attribute won't work.

Oh yeah.. whoops! :)

davidwengier:
Something I realised that you might want to double check in your test @JamesNK: Does the `// language=regex` work if it is immediately follow by line mappings? 

eg:

```
[...]

PR #367, opened 2022-10-15
Add route string syntax to Razor pages and Blazor

JamesNK:
Fixes https://github.com/dotnet/razor-compiler/issues/352

I've hacked this in. There are things todo:
1. Is this the right way to detect whether a Razor document is a Razor Page or Blazor page?
2. What is the simplest way for me to test this in Visual Studio? I'm not sure whether VS will detect and use the string syntax in the directive. Follow up work here, or in Roslyn, might be required.
3. Update dozens of failing unit tests. I'm guessing you don't update the documents manually. Is there a way to regenerate all the documents?

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/src/CodeGeneration/DefaultCodeRenderingContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/src/CodeGeneration/DefaultCodeRenderingContext.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 #nullable disable
@@ -90,7 +90,7 @@ internal class DefaultCodeRenderingContext : CodeRenderingContext
 
     public override RazorDiagnosticCollection Diagnostics { get; }
 
-    public override string DocumentKind { get; }
+    public override string DocumentKind => _documentNode.DocumentKind;
 
     public override ItemCollection Items { get; }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Language/src/CodeGeneration/DefaultCodeRenderingContext.cs
```diff
@@ -90,7 +90,7 @@ internal class DefaultCodeRenderingContext : CodeRenderingContext
 
     public override RazorDiagnosticCollection Diagnostics { get; }
 
-    public override string DocumentKind { get; }
+    public override string DocumentKind => _documentNode.DocumentKind;
```
JamesNK, line 93: DocumentKind is never set for some reason 🤷

333fred:
@JamesNK this repo is effectively in servicing only, as we're moving to combine with the razor-tooling repo. We're only accepting servicing fixing for 7.0.1xx here (@jaredpar for a bar check, but I don't think this makes the bar). The code has been moved to `dotnet/razor-tooling/src/Compiler`. You'll also find it much easier to test your changes in VS: you should just be able to F5 the tooling repo and see your compiler changes reflected live in the debug instance (which was one of several reasons for this change :)).

333fred:
As to updating the baselines, you can do a `dotnet test` run with the `GENERATE_BASELINES` environment variable set to `true`.

JamesNK:
Ok. You should update the readme - https://github.com/dotnet/razor-compiler/blob/main/README.md - or repo description to make that obvious.

333fred:
We're still in the process of it. We'll have our ducks in a row soon.

